PR #144, opened 2020-02-28
namespace converted to object since babel does not play well with this

kutlugsahin:
Hey Ferdi,

I was playing with this in a next app. Babel complained about this namespaced non-const exports so I updated it to an object and can be used with the same syntax. And again good job! Very useful library  :)

Kutlu

### DIFF
--- a/src/templates/core/OpenAPI.hbs
+++ b/src/templates/core/OpenAPI.hbs
@@ -3,9 +3,9 @@
 /* eslint-disable */
 /* prettier-ignore */
 
-export namespace OpenAPI {
-    export let BASE = '{{{server}}}';
-    export let VERSION = '{{{version}}}';
-    export let CLIENT = '{{{httpClient}}}';
-    export let TOKEN = '';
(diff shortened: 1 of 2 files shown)


ferdikoomen:
Hey @kutlugsahin I think there is an issue with this object notation: If somewhere in your app's code you import this object and change a parameter: OpenAPI.TOKEN for instance then the 'original' reference is not changed.

Plus.. all models and enums are also still wrapped inside namespaces? Or is Babel fine with that?

kutlugsahin:
Babel is complaining **only if** namespace notation is used to export **non-const** properties as in the OpenAPI namespace. I guess thats why it does not complain about the other namespaces. For exporting object; it's fine to export as object and all imports will get the same reference since modules are basically singletons. Cheers!!

ferdikoomen:
Hi Kutlu, i made the change in another branch, took your approach, this is merged now ,im testing it today to see if it has no real impact.

PR #151, opened 2020-03-06
propertyRequired with definition.required === true

ismail-codar:


ferdikoomen:
Hey Ismail, let me check why the build failed.

ferdikoomen:
According to the spec required is an array of props, rather then a boolean i believe

ferdikoomen:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject

ismail-codar:
My schema object like this.
```json
"schemas": {
      "ErrorResponse": {
        "properties": {
          "type": {
            "type": "string",
            "required": true
```
It looks out of the spec.  I need to contact the API provider

PR #152, opened 2020-03-06
propertyRequired with definition.required === true

ismail-codar:


ismail-codar:
https://github.com/ferdikoomen/openapi-typescript-codegen/pull/151

Issue #249, opened 2020-06-04
Feature request: Option to change how cookies are sent with requests


gazab:
To allow us to use cookie based authentication and make the browser pass cookies to the server on each request we need to tell fetch to allow it. This is done by setting _credentials_ to "same-origin" or "include" when doing requests. Right now this is hardcoded to "same-origin" which only works if the client and API has the same origin. 

I would like to set it to "include" by passing some option so I can have the client and API on different origins and still use cookie authentication.

ferdikoomen:
Good point @gazab Do you want to set this at the moment of generation? Otherwise im thinking we can add this to the OpenApi global settings (where we also set the bearer token for instace)

gazab:
@ferdikoomen Hm, I think I would prefer the global options for anything not required to be done on generation. And the old value should still be default. What do you think?

ferdikoomen:
Sounds good, adding this now

ferdikoomen:
Created PR: https://github.com/ferdikoomen/openapi-typescript-codegen/pull/256

OpenApi.WITH_CREDENTIALS = true

Will set the correct settings for both fetch and XHR request

ferdikoomen:
Will create a new build in a bit, PR is merged

PR #256, opened 2020-06-11
- Added with credentials

ferdikoomen:
Should fix issue #249


gazab:
Nice catch with XHR equivalent! Kinda forgot about XHR entirely :)

PR #296, opened 2020-07-23
Replace 'export' with 'export type' for models

iegik:
...which is TypeScript interfaces and should be exported differently

### DIFF
--- a/src/templates/index.hbs
+++ b/src/templates/index.hbs
@@ -12,7 +12,7 @@ export { OpenAPI } from './core/OpenAPI';
 {{#if models}}
 
 {{#each models}}
-export { {{{this}}} } from './models/{{{this}}}';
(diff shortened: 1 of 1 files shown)


ferdikoomen:
@iegik We cannot just export 'type' here, when using Enum's the models actually contain a definition like this:

```
export interface Pet {
    id?: number;
    name: string;
    category?: Category;
    photoUrls: Array<string>;
    tags?: Array<Tag>;
    status?: Pet.status;
}
export namespace Pet {
    export enum status {
        AVAILABLE = 'available',
        PENDING = 'pending',
        SOLD = 'sold',
    }
}
```
Otherwise i would have export the models a **.d.ts** files (typescript definition files). In big projects these tend to compile faster. But due to the enums we cannot.

PR #377, opened 2020-10-05
Allow asynchronous bearer token

### LINKED ISSUE #376: Fire promise globally before any request
First off, thanks so much for this library. This is really helpful.

# Problem

I am currently using the `OpenAPI.TOKEN` syntax to set my auth token. This works well, but it isn't perfect for my use case. The reason is that my token is accessed asynchronously.

Ideally, I could wrap the `fetch` function used by this lib, such that it gets my token asynchronously before every call.

I'm using Firebase auth, which 1) refreshes the token often, and 2) has an asynchronous function to get the current token.

Currently, I do this:

```js
onAuthStateChanged(async (user) => {
  if (user) {
   // the user is set as signed in here, so I update the app's auth state
   setIsSignedIn(true) // <- this is pseudo-code. This is what Firebase does for me, for illustration.
   // thus, they're now seeing screens that require an auth token

  // it's possible to be stuck here, where the app state has updated to show authenticated screens
  // but the token promise hasn't resolved yet

   // next, I set the user's token for request
   OpenAPI.TOKEN = await user.getIdToken() 
  } else {
    OpenAPI.TOKEN = ''
  }
})
```

The problem with this is that `await user.getIdToken()` is async. Since I render authenticated screens right when `user` exists, then it's possible that the user will see certain screens before I have set the `OpenAPI.TOKEN`. This means requests that should have a token will be sent without one.

# Proposed solution

At the root of my app, it would nice to be able to do something like this:

```ts
import { OpenAPI } from 'openapi-typescript-codegen'

OpenAPI.onBeforeRequest = async () => {
   OpenAPI.TOKEN = await firebase.auth().currentUser?.getIdToken() 
}
```

This way, before any request is sent, I know with certainty that the latest auth token is injected in my requests.

## My current work-around

Currently, I have to manually call this before every request:

```ts
import { SomeService } from '../generated'

export const callServer = async () => {
   OpenAPI.TOKEN = await firebase.auth().currentUser?.getIdToken() 
   return SomeService.someCall()
}
```

Doing this manually isn't ideal, since I might forget some.

I'm definitely open to a better solution, and am happy to explain better if this is unclear. Thank you!

nandorojo:
This PR closes #376. 

Sometimes, you need to run an asynchronous function to get your token from your auth provider. In order to always have the most up-to-date token, this function should be called at the time of request.

It seems like the tests pass for this, I'm testing it in my project now too. Will report back.

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,6 +339,20 @@ import { OpenAPI } from './generated';
 OpenAPI.TOKEN = 'some-bearer-token';
 ```
 
+Alternatively, we also support an async method that provides the token for each request.
+You can simply assign this method to the same `TOKEN `property in the global OpenAPI object.
+
+```typescript
+import { OpenAPI } from './generated';
+
+const getToken = async () => {
+    // Some code that requests a token...
+    return 'SOME_TOKEN';
+}
+
+OpenAPI.TOKEN = getToken;
+```
+
 
 ### Compare to other generators
 Depending on which swagger generator you use, you will see different output.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-    "name": "openapi-typescript-codegen",
-    "version": "0.5.0",
+    "name": "@beatgig/openapi-typescript-codegen",
+    "version": "0.5.0-async-token.2",
     "description": "NodeJS library that generates Typescript or Javascript clients based on the OpenAPI specification.",
     "author": "Ferdi Koomen",
     "homepage": "https://github.com/ferdikoomen/openapi-typescript-codegen",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-    "name": "@beatgig/openapi-typescript-codegen",
-    "version": "0.5.0-async-token.2",
+    "name": "openapi-typescript-codegen",
+    "version": "0.5.1",
     "description": "NodeJS library that generates Typescript or Javascript clients based on the OpenAPI specification.",
     "author": "Ferdi Koomen",
     "homepage": "https://github.com/ferdikoomen/openapi-typescript-codegen",
--- a/src/templates/core/OpenAPI.hbs
+++ b/src/templates/core/OpenAPI.hbs
@@ -4,7 +4,7 @@ interface Config {
     BASE: string;
     VERSION: string;
     WITH_CREDENTIALS: boolean;
-    TOKEN: string | () => Promise<string>;
+    TOKEN: string | (() => Promise<string>);
 }
 
 export const OpenAPI: Config = {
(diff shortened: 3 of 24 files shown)


nandorojo:
I updated a few types and fixed it up. It's working well in my app.

ferdikoomen:
Hi @nandorojo Although im not sure if this async token handler is a very good idea (seems like logic that the application should handle), but the change is pretty small :-) This does mean that for each call the OpenAPI client will get / wait for a new bearer token.., that seems a bit weird. The idea with these bearer tokens is that they expire after some time and you need to re-fetch them at that point. Normally you would have some kind of client (like the OpenID client) that would have some kind of event that gives you a new token, just before the last one expires:

```
const userManager = new UserManager({...});
userManager.events.addUserLoaded(user => {
    OpenAPI.TOKEN = user.access_token;
});
```
This is an example from the OpenID client: https://github.com/IdentityModel/oidc-client-js/wiki. I assume that Firebase auth also has a very similar concept (onTokenRefresh): https://firebase.google.com/docs/reference/js/firebase.messaging.Messaging#ontokenrefresh

nandorojo:
@ferdikoomen thanks for getting back to me!

I understand what you mean, but without doing this, requests won't reliably use the correct token in each call. And it feels like a much simpler approach than a long work-around: just get the token before making the call.

In the case of Firebase, the function is asynchronous, but if the token hasn't expired, it returns the current token "synchronously". Firebase only makes an asynchronous request if the token is past the expiration time to refresh it. This means that in almost every request, the token getter function is simply returning the current token without any async requests. But in the off chance that it needs to refresh, I wouldn't want to send a request with a stale token.

The alternative is, I refactor my entire app and use something like Redux to track the token every time it changes, and conditionally render pages based on this token. But this could lead to flashes of empty content when a token refreshes. I would rather implement this logic in my network requests, and I agree that this library should handle it. Plus, the behavior is fully opt-in, and not breaking at all :)

I've been using it in my app, and it works really well.

nandorojo:
Just to follow up, Firebase does have this functionality: https://firebase.google.com/docs/reference/js/firebase.auth.Auth

However, the `user` it returns doesn't directly return a `token` field, but rather an `async` `getIdToken()` function.

nandorojo:
@ferdikoomen Just wanted to see if you're open to merging this. If not, I'll sadly have to work off my own fork. Would you be open to at the very least merging it undocumented?

To reiterate what I mentioned before: my token getter function is `async`, but almost every time it's called, it's returning a cached value synchronously. However, in order to keep it perfectly up-to-date with refresh tokens, I need to check for the token before each network request. The best solution for this is to allow the `TOKEN` to be a function that returns a string.

Thanks again for the great work on this library. We're only a team of 2 (one front-end and one back-end) and it's made our workflow way easier.

ferdikoomen:
@nandorojo I added some changes to the PR, with support for the XHR flow, some added documentation and the E2E tests. If the tests pass then ill merge the PR!

nandorojo:
@ferdikoomen  thanks so much!

PR #399, opened 2020-10-28
Implement support for anonymous models described using anyOf or oneOf…

thetrime:
…. allOf seems OK. Remove TODO. Add tests. Credit Yusei Abe in LICENSE for the core idea of the code, found at https://github.com/diverta/kuroco-sdk/blob/8029698bc6d43b94a9190e7d1f34dad92234b3ed/src/generator/openApi/v3/parser/getModel.ts#L120

This is a bit simple than the approach taken by @yabe-diverta in that it does not handle nested anyOf and allOf (though as far as I can tell, yabe-diverta's code only handles one more level of nesting than mine, rather than arbitrary nesting). The tests all still pass, and I've added two new ones for the anyOf and oneOf cases described.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ junit.xml
 .idea
 .vscode
 *.iml
-dist
 coverage
 test/generated
 test/e2e/generated
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,7 @@
 MIT License
 
 Copyright (c) 2020 Ferdi Koomen
+Copyright (c) 2020 Yusei Abe
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/dist/index.js
+++ b/dist/index.js
@@ -3108,26 +3108,26 @@ var templateExportService = {"1":function(container,depth0,helpers,partials,data
     + ((stack1 = lookupProperty(helpers,"if").call(alias1,lookupProperty(depth0,"deprecated"),{"name":"if","hash":{},"fn":container.program(7, data, 0),"inverse":container.noop,"data":data,"loc":{"start":{"line":17,"column":4},"end":{"line":19,"column":11}}})) != null ? stack1 : "")
     + ((stack1 = lookupProperty(helpers,"if").call(alias1,lookupProperty(depth0,"summary"),{"name":"if","hash":{},"fn":container.program(9, data, 0),"inverse":container.noop,"data":data,"loc":{"start":{"line":20,"column":4},"end":{"line":22,"column":11}}})) != null ? stack1 : "")
     + ((stack1 = lookupProperty(helpers,"if").call(alias1,lookupProperty(depth0,"description"),{"name":"if","hash":{},"fn":container.program(11, data, 0),"inverse":container.noop,"data":data,"loc":{"start":{"line":23,"column":4},"end":{"line":25,"column":11}}})) != null ? stack1 : "")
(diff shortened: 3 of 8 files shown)


yabe-diverta:
thanks for the mention @thetrime .

I think it's not necessary that the modified copyright that includes me.
Because my codes are just patches, they are imperfect.
**it may cause some malfunctions in some cases as well**.

as a background story,
at the time of launching a new PJ in my team,
they planned to expose it's backend definition as OpenAPI V3,
and i was assigned to make a SDK to handle it easily on front-end (JS/TS) side.

at the first time of first-developing term,
i thought to just use this good tool as a library as one of the dependencies of my SDK.
I just wanted a simple tool that can perse OpenAPI V3 definition to a TS code.
althogh i could not fix the mulfunctions of OneOf... one.
so that i needed to hack and reproduce this repo to override some codes as patches.
i would switch to just add this tool as a dependency of my SDK instead of hacking if the probrem ware fixed.

the backend of my team generates OpenAPI definitions in a very few pattern,
i did a few tests to just fulfill those particular patterns.
i don't guarantee that this patch can be appropriate for general patterns.
I guess it's required to redesign and refine from it's root level to solve this OneOf... one.

Anyway, i don't recommend to merge my patch (this pull request seems to be more sophisticated then mine?).
i would contribute to solve this (if i have much time...).
thanks @thetrime @ferdikoomen (@artem-tim)

budde377:
I suspect that the implemented logic could be extended to the `allOf` case. Which would solve cases like the following:

```json
...
"ModelWithNullableReferenceProperty": {
    "description": "This is a model which property is a reference and nullable",
    "type": "object",
    "properties": {
        "nullableProp": {
            "allOf": [
                {
                    "$ref": "#/components/schemas/ModelWithString"
                },
                {
                     "$ref": "#/components/schemas/ModelThatExtends"
                }
            ],
            "nullable": true
        }
    }
},
...
```

which currently is rendered as
```typescript
export interface ModelWithNullableReferenceProperty {
    nullableProp?: any;
}
```

but with the same logic as above this, could be rendered as

```typescript
export interface ModelWithNullableReferenceProperty {
    nullableProp?: ModelWithString & ModelThatExtends | null;
}
```

thetrime:
> thanks for the mention @thetrime .
> 
> I think it's not necessary that the modified copyright that includes me.
> Because my codes are just patches, they are imperfect.

All code is imperfect :) If you don't want to be mentioned, I can certainly remove it.

> **it may cause some malfunctions in some cases as well**.

Yes, I'm aware this is a bit rough, but it's better than what was available before. If someone wants to refine it even further, then that's fine by me.

> as a background story,
> at the time of launching a new PJ in my team,
> they planned to expose it's backend definition as OpenAPI V3,
> and i was assigned to make a SDK to handle it easily on front-end (JS/TS) side.
> 
> at the first time of first-developing term,
> i thought to just use this good tool as a library as one of the dependencies of my SDK.
> I just wanted a simple tool that can perse OpenAPI V3 definition to a TS code.
> althogh i could not fix the mulfunctions of OneOf... one.
> so that i needed to hack and reproduce this repo to override some codes as patches.
> i would switch to just add this tool as a dependency of my SDK instead of hacking if the probrem ware fixed.

Yes - I'm in exactly the same situation.

> the backend of my team generates OpenAPI definitions in a very few pattern,
> i did a few tests to just fulfill those particular patterns.
> i don't guarantee that this patch can be appropriate for general patterns.
> I guess it's required to redesign and refine from it's root level to solve this OneOf... one.

It works for my patterns too. Note that my code is even *more* limited than your patch - it won't handle nested oneOf properly at all

budde377:
The PR #419 solves this, but with objects of arbitrary depth and is not limited to interfaces.

ferdikoomen:
@budde377 @thetrime  This PR works, but the code is hard to read. However it works, so im tempted to merge it for now and look at this story a bit better in the future.

ferdikoomen:
Closing this in favour of https://github.com/ferdikoomen/openapi-typescript-codegen/pull/419

Issue #400, opened 2020-10-28
Support for custom request-function


nils4cosee:
**Describe the solution you'd like**
In our last project, we have used a custom axios setup to access the backend. One thing that would not have been possible with "openapi-typescript-codegen" is the axios-interceptor that retrieves the the current language settings from "i18next" and adds a query parameter to the outgoing request. It was very convenient to do it this way, because from that point on, we didn't have to worry about transmitting the correct language anymore.

I would like to see a mechanism that uses a custom request-function. That way, everybody can write an adapter if needed.

The following changes 

```typescript
// OpenAPI.ts
export type RequestFunction = (options: ApiRequestOptions) => Promise<ApiResult>;

interface Config {
    BASE: string;
    VERSION: string;
    WITH_CREDENTIALS: boolean;
    TOKEN: string | (() => Promise<string>);
    requestFunction?: RequestFunction;  // added
}

// request.ts
export async function request(options: ApiRequestOptions): Promise<ApiResult> {
     if (OpenAPI.requestfunction != null) {
         return OpenAPI.requestFunction(options);
     }
     ...
}

```

would allow a configuration similar to the bearer-token:

```typescript

OpenAPI.requestFunction = async (options) => {
    ...
    return {
        ...
    }
}
```

I briefly thought about using your library in our project, but it was too late by then. In the next project, maybe. It could be a huge time-saver.

ferdikoomen:
@nils4cosee The reason there is no 'requestFunction' is because the generator support fetch and XHR and both have quite different API options. The reason for supporting only these options is because this needs to be a lightweight 'no-libraries / no-dependencies' client. I get the idea of changing internals or creating these 'callbacks' (its been requested before), but for these advanced use cases you are better of with the JAVA based openapi-codegen. 

But let me look at it from a different view: How would your SPEC look like if you always need this 'language' property? Would there be this prop for each of the service calls?

nils4cosee:
I have had a look at the swagger-spec of our API again. The query-param is currently not part of all endpoints. This is mainly, because we auto-generate the spec from the Spring-Boot endpoints and the query-parameter is processed in a Filter before the Controller classes are invoked. Since the parameter is universal, I would like to treat it in a universal fashion in the frontend as well. And I really don't want to pull any Java-dependencies into the frontend project..

[...]

PR #407, opened 2020-11-02
commit to illustrate changes for custom-http-client (do not merge)

nils4cosee:
This example illustrates the feature-request in #400 

The library-user can specify a custom function that sends requests to the backend (as illustrated in `test/e2e/v3.custom.spec.js`

Helper functions like "getToken" and "getUrl" are not provided and must be written by the library-user.

The downside is that `ApiResult` and `ApiRequestOptions` would become part of the official API, which will impose restrictions in SemVer-versioning. I can understand if you don't want that.

Final polishing and tests are still missing. Before I go any further I would like to hear your opinion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ $ openapi --help
     -V, --version             output the version number
     -i, --input <value>       OpenAPI specification, can be a path, url or string content (required)
     -o, --output <value>      Output directory (required)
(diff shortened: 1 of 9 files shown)


ferdikoomen:
Closing this PR in favor of https://github.com/ferdikoomen/openapi-typescript-codegen/pull/434 and already existing option to overwrite request.ts

PR #419, opened 2020-11-10
feat: Improve support for allOf, oneOf, and anyOf

budde377:
The current composition schemas `allOf`, `oneOf`, and `anyOf`, only works on references. In this PR I've added support for types of arbitrary depth and type. This is done by promoting three composition types on the `Model.export` to preserve type information for the subsequent schema and model generation. 

I.e. the `Model` now have the type

```typescript
type Model = {
  export: 
    | 'reference' 
    | 'generic' 
    | 'enum' 
    | 'array' 
    | 'dictionary' 
    | 'interface' 
    | 'one-of' 
    | 'any-of' 
    | 'all-of'
  type: ...
}
```

### Types

The following specification:
```json
{
  "type": "object",
  "properties": {
    "any": {
      "anyOf": [
        { "$ref": "#/components/schemas/A" },
        { "$ref": "#/components/schemas/B" },
        { "$ref": "#/components/schemas/C" },
      ]
    },
    "one": {
      "oneOf": [
        { "$ref": "#/components/schemas/A" },
        { "$ref": "#/components/schemas/B" },
        { "$ref": "#/components/schemas/C" },
      ]
    },
    "all": {
      "allOf": [
        { "$ref": "#/components/schemas/A" },
        { "$ref": "#/components/schemas/B" },
        { "$ref": "#/components/schemas/C" },
      ]
    }
  }
}
```

will render the type

```typescript
... {
  any: A | B | C
  one: A | B | C // Notice: TypeScript doesn't (easily) support one-of, so this is the closest I could get for now.
  all: A & B & C
}
```

This approach will also, easily, support nullable references (which was my original motivation):

```json
{
  "type": "object",
  "properties": {
    "all": {
      "allOf": [
        { "$ref": "#/components/schemas/A" },
        { "$ref": "#/components/schemas/B" },
        { "$ref": "#/components/schemas/C" },
      ],
    "nullable": true
    }
}
```

will render the type

```typescript
... {
  all: (A & B & C) | null
}

```

###  Schema

I wasn't sure what the preferred way of expressing this in the schema was. The current implementation changes the specification

```json
{
  "type": "object",
  "properties": {
    "any": {
      "anyOf": [
        { "$ref": "#/components/schemas/A" },
        { "$ref": "#/components/schemas/B" },
        { "$ref": "#/components/schemas/C" },
      ]
    },
    "one": {
      "oneOf": [
        { "$ref": "#/components/schemas/A" },
        { "$ref": "#/components/schemas/B" },
        { "$ref": "#/components/schemas/C" },
      ]
    },
    "all": {
      "allOf": [
        { "$ref": "#/components/schemas/A" },
        { "$ref": "#/components/schemas/B" },
        { "$ref": "#/components/schemas/C" },
      ]
    }
  }
}
```
to

```
.. {
  properties: {
    one: { type: 'one-of', contains: [ { /* A */ }, { /* B */ }, { /* C */ } ]  },
    any: { type: 'any-of', contains: [ { /* A */ }, { /* B */ }, { /* C */ } ]  },
    all: { type: 'all-of', contains: [ { /* A */ }, { /* B */ }, { /* C */ } ]  },
  }
}
```

### DIFF
--- a/src/client/interfaces/Model.d.ts
+++ b/src/client/interfaces/Model.d.ts
@@ -3,7 +3,7 @@ import type { Schema } from './Schema';
 
 export interface Model extends Schema {
     name: string;
-    export: 'reference' | 'generic' | 'enum' | 'array' | 'dictionary' | 'interface';
+    export: 'reference' | 'generic' | 'enum' | 'array' | 'dictionary' | 'interface' | 'one-of' | 'any-of' | 'all-of';
     type: string;
     base: string;
     template: string | null;
--- a/src/openApi/v3/parser/getModel.ts
+++ b/src/openApi/v3/parser/getModel.ts
@@ -125,70 +125,42 @@ export function getModel(openApi: OpenApi, definition: OpenApiSchema, isDefiniti
             return model;
         }
     }
+    // TODO:
+    //  Add correct support for oneOf
+    //  https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/
 
-    if (definition.anyOf?.length && !definition.properties) {
-        const compositionTypes = definition.anyOf.filter(type => type.$ref).map(type => getType(type.$ref));
-        const composition = compositionTypes
-            .map(type => type.type)
-            .sort()
-            .join(' | ');
-        model.export = 'generic';
-        model.imports.push(...compositionTypes.map(type => type.base));
-        model.type = composition;
-        model.base = composition;
+    if (definition.oneOf?.length || definition.anyOf?.length || definition.allOf?.length) {
+        let types: OpenApiSchema[] = [];
+        if (definition.oneOf?.length) {
+            model.export = 'one-of';
+            types = definition.oneOf;
+        } else if (definition.anyOf?.length) {
+            model.export = 'any-of';
+            types = definition.anyOf;
+        } else if (definition.allOf?.length) {
+            model.export = 'all-of';
+            types = definition.allOf;
+        }
+        const compositionTypes = types.map(model => getModel(openApi, model));
+        model.properties = compositionTypes;
+        model.imports.push(...compositionTypes.reduce((acc: string[], type) => acc.concat(type.imports), []));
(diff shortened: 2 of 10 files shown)


budde377:
@ferdikoomen Not trying to be pushy here, but WDYT?

ferdikoomen:
@budde377 this PR generated the types more correct, however i see two code issues:

The **ModelThatExtends** generates the following output:
```
export type ModelThatExtends = ( & ModelWithString & {
    propExtendsA?: string,
    propExtendsB?: ModelWithString,
});
```
The first `&` is incorrect, this should be:
```
export type ModelThatExtends = (ModelWithString & {
    propExtendsA?: string,
    propExtendsB?: ModelWithString,
});
```
Similar with **ModelWithAnyOfAndNullable**, this generates:
```
export interface ModelWithAnyOfAndNullable {
    propA?: ( | {
        boolean?: boolean,
    } | ModelWithEnum | ModelWithArray | ModelWithDictionary) | null;
}
```
The first `| {` is incorrect, this should be:
```
export interface ModelWithAnyOfAndNullable {
    propA?: ({
        boolean?: boolean,
    } | ModelWithEnum | ModelWithArray | ModelWithDictionary) | null;
}
```

ferdikoomen:
but it looks promising, if the small issues are resolved we should be able to merge it in favour of #399

budde377:
@ferdikoomen I've updated the code to not prepend pipe and ampersand. While the generated types were valid, as you may write

``` typescript
type Q = A | B | C
```
as
```typescript
type Q = 
  | A
  | B 
  | C 
```
the updated generated code is definitely causing less confusion.

ferdikoomen:
@budde377 i aligned it with the latest changes from the master, plus i used the "unless last" that we use in other areas as well (to align). I think the result looks good. Let me create a 0.7.0-beta package out of this and test it in production.

ferdikoomen (approved):

PR #433, opened 2020-11-19
[incomplete] Allow external references

chris-armstrong:
_I'm opening this PR to solicit feedback_

At the moment, openapi-typescript-codegen can only resolve `$ref` references to components in the same file. I've begun **working on support for resolving external `$ref` references** (i.e. where the openapi file has '$ref' entries that refer to other files relative to it on the filesystem).

This makes it easier to split a large openapi.yml into multiple files. It also helps to bring the tool closer in line with the [JSON Reference spec](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03), which is allowed by the [OpenAPI 3.0 specification](https://swagger.io/specification/#relative-references-in-urls). 

_I've deliberately neglected URL references other than those on the current filesystem for now, but there's no reason this couldn't be added (perhaps behind a flag for those who don't want it making HTTP calls?)_

This PR isn't complete, as in not all cases where `$ref` is allowed have been updated and tested, and it has only been done for the 3.0 side.

Could I get some feedback as to whether this PR would be accepted if continued along its current lines, or if it will require some more significant changes to the way it is structured? I'd love to be able to get this functionality into the main codebase.


ferdikoomen:
@chris-armstrong Interesting Chris! I'm working on some core rewrites, the resolving / bundeling logic is indeed nice to add.

chris-armstrong:
> @chris-armstrong Interesting Chris! I'm working on some core rewrites, the resolving / bundeling logic is indeed nice to add.

Great! I'll continue working on it to clean it up and get it tested properly and add it back to the `v2` schema parsing / generation.

ibolmo:
@ferdikoomen this is something I need. You had mentioned some core rewrites. Would you say that is blocking this PR from moving forward?

chris-armstrong:
Instead of continuing development of this one, which was becoming extensive, I've found a library that achieves the same effect and can be integrated with minimal code change, so I've opened #573 instead to achieve the same outcome.

PR #448, opened 2020-11-26
Add support for enum names with spaces inside them.

mlaps-gafe:
If enum names have spaces inside them, then the library generates a broken output, something like that:

export enum MyCustomEnum {
  WithoutSpaces = 1,
  With Spaces = 2, <--- error here
}

After the changes introduced in this PR, the output will look like that:

export enum MyCustomEnum {
  WithoutSpaces = 1,
  "With Spaces" = 2, <--- no error here
}


ferdikoomen:
@mlaps-gafe Does this happen when using the "x-enum-varnames" extension?

mlaps-gafe:
> @mlaps-gafe Does this happen when using the "x-enum-varnames" extension?

Exactly. When I put there a string that contains spaces.

mlaps-gafe:
@ferdikoomen If you are ok with the change, how long will it take to get it merged in?

PR #489, opened 2020-12-15
Add custom fetcher as api config parameter

serhiiminin:
There is an ability to provide a custom fetcher in case you need manual control over it


serhiiminin:
Agreed on the other solution

Issue #429, opened 2020-11-18
Ability to set "global" headers on client generation


Alex0007:
![image](https://user-images.githubusercontent.com/1143226/99484399-25b46a00-2982-11eb-9baf-4ba300cc5f67.png)

I want to specify some "global" headers to pass package version to API (for example `X-Client` header)
I have [implemented](https://github.com/Alex0007/openapi-typescript-codegen/commit/fb6d48bda9ad0509d71f68b3c44927c9d48a9c9c) this functionality in my fork (but now it's too outdated comparing to current master branch)

Current `openapi-typescript-codegen` situation: headers can be set when using generated code only (not when generating client).

Maybe idea isn't that bad and can be added into this project?

ferdikoomen:
@Alex0007 closing this one. Recently we had some similar requirements and it's easy to set these headers in runtime, or alternatively publish a package from the generated client with those headers set.

Issue #428, opened 2020-11-17
Use without global variables


cobraz:
Hello,

I realize after using this library that using global variables is suboptimal for what I'm using it for.

In my case, I have a Node.js application. I need to run multiple processes with different auth (API tokens) for each, and having a global scope would limit my performance quite a bit (i have to run all in sequence to make sure that the right token is used for the request).

For quite some time, I used another library to generate code. However, that had classes for each service (like yours), but not with singletons. You had to construct each class and pass auth. E.g. `new PostService('token')`.

Code that handled more than one service would have to construct a few classes. Also sub-optimal. IMO, for the best possible API, I'd like to have one class that reference others.

Example:

```typescript
class MyApi {

	private auth;

	constructor(... some auth thing ...) {}

	postService() {
		return new PostService(auth);
	}
}
```

In summary, I'd love to see a way where this library could a) generate classes without singletons and b) have a class that exports constructed services.

I realize looking at your code that this isn't much work to do. Would you accept a pull request with that? I could also fork, but I would much rather help out :)

cobraz:
Actually, you could probably introduce some configuration, not just auth, as such:

```
class MyApi {

	private config;

	constructor(... some config ...) {}

	postService() {
		return new postService(auth);
	}
}
```

ferdikoomen:
@cobraz If you generate two OpenAPI clients and use a default (webpack, parcel, rollup) bundler then each client can have its own module scope. I will attach an example of the setup in the ticket. This is a much neater setup then exposing classes for this, otherwise you will need to pass around this MyApi class everywhere.

vitalybaev:
Another good reason not to use global variables is SSR - using global variables to update token can cause issues when some user could occasionally become logged in as another user.

ferdikoomen:
@vitalybaev @cobraz There is a new version (0.16.0) that allows to create client instances to each client can have its own bearer token. You can specify the `--name` argument while generating the client. More info:
https://github.com/ferdikoomen/openapi-typescript-codegen#generate-client-instance-with---name-option

mb21:
For us, the `--name` option was not flexible enough, so we let the user pass in a whole template for the services: https://github.com/ferdikoomen/openapi-typescript-codegen/pull/1268

silasdavis:
> otherwise you will need to pass around this MyApi class everywhere.

[...]

PR #492, opened 2020-12-17
Add 'injected' HTTP client option

### LINKED ISSUE #400: Support for custom request-function

### LINKED ISSUE #465: Option to dependency inject the request function
I would like to have an option to dependency inject the request function in each operation, such that the services don't have a dependency on the core modules, and so that `--exportCore false` will produce typescript code that compiles.

For example:
```
type __request = (options: ApiRequestOptions) => Promise<ApiResult>

export class MyService {
    public static async myOperation(
        request: __request,
        myParam: string,
        requestBody: MyRequestModel,
    ): Promise<MyResponseModel> {
        const result = await request({
            method: 'POST',
            path: `/my-api/${myParam}`,
            body: requestBody,
        });
        return result.body;
    }
}
```

Use case 1:

I consume multiple OpenAPIs from multiple different microservices. I need to make modifications to the core request module to make it support React Native. I need to make modifications to the core OpenAPI module to fetch the correct token. I want to maintain one copy of these modifications across the apis generated for all the microservices, instead of duplicating the modifications per microservice.

Use case 2:

I modified the core request module to add a timeout. I want to use different timeouts in different operation calls. If the request function was given as a parameter to the operation, I could pass a modified request function that included the timeout.

Example:
```
const r = await MyService.myOperation(makeRequest({timeout: 10}), "string", {object: true})
```

May be related to #428 and #400

Thanks for a great library.

iansan5653:
Adds a new HTTP client option `injected` which will not generate an HTTP client and will instead add a field to all service classes in which the consumer supplies a request method in the form `(options: ApiRequestOptions) => Promise<any>`. This would be the most flexible option available, as now anyone would have the power to use any request tools available. They can track and log requests, manage their own authentication, customize behavior to fit their exact needs, and inject a mock/spy method for testing.

Additional changes made to facilitate this:

- The standard `request` method exported by all HTTP clients now returns the body directly (`Promise<any>`) instead of an object containing the body (`Promise<ApiResponse>`). This makes it easier to create a method as the injected method won't have to return unnecessary information.
- All services are now standard classes that take the request method as a constructor argument, regardless of which HTTP client is being used. This would be a breaking change, except that when the HTTP client is not `injected`, the services are exported as pre-instantiated classes already bound to the HTTP client. So in that case, this change is invisible.
- Added a condition so `request.ts` is not generated if HTTP client is `injected`.
- Updated tests and snapshots.

Resolves issues:
* Resolves #465 
* Potentially #429 
* Potentially #428 
* Resolves #400 

Also may replace / affect PRs #407, #434, #489.

TODO:
- [ ] Update Readme with documentation for new option and how to use

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,6 +11,7 @@ export enum HttpClient {
     FETCH = 'fetch',
     XHR = 'xhr',
     NODE = 'node',
+    INJECTED = 'injected'
 }
 
 export type Options = {
--- a/src/templates/core/fetch/request.hbs
+++ b/src/templates/core/fetch/request.hbs
@@ -50,10 +50,10 @@ import { OpenAPI } from './OpenAPI';
 /**
  * Request using fetch client
  * @param options The request options from the the service
- * @returns ApiResult
+ * @returns The response from the API
  * @throws ApiError
  */
-export async function request(options: ApiRequestOptions): Promise<ApiResult> {
+export async function request(options: ApiRequestOptions): Promise<any> {
     const url = getUrl(options);
     const response = await sendRequest(options, url);
     const responseBody = await getResponseBody(response);
@@ -68,5 +68,5 @@ export async function request(options: ApiRequestOptions): Promise<ApiResult> {
     };
 
     catchErrors(options, result);
-    return result;
+    return result.body;
 }
--- a/src/templates/core/node/request.hbs
+++ b/src/templates/core/node/request.hbs
@@ -54,10 +54,10 @@ import { OpenAPI } from './OpenAPI';
 /**
  * Request using node-fetch client
  * @param options The request options from the the service
- * @returns ApiResult
+ * @returns The response from the API
  * @throws ApiError
  */
-export async function request(options: ApiRequestOptions): Promise<ApiResult> {
+export async function request(options: ApiRequestOptions): Promise<any> {
     const url = getUrl(options);
     const response = await sendRequest(options, url);
     const responseBody = await getResponseBody(response);
@@ -72,5 +72,5 @@ export async function request(options: ApiRequestOptions): Promise<ApiResult> {
     };
 
     catchErrors(options, result);
-    return result;
+    return result.body;
 }
--- a/src/templates/core/xhr/request.hbs
+++ b/src/templates/core/xhr/request.hbs
@@ -53,10 +53,10 @@ import { OpenAPI } from './OpenAPI';
 /**
  * Request using XHR client
  * @param options The request options from the the service
- * @returns ApiResult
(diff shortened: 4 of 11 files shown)


iansan5653:
It looks like CI is failing because the snapshots have different indentation when generated on my machine. I also noticed some missing indentation in the output for a project I'm working on, so I think that may be a separate unrelated bug - maybe related to using Windows?

iansan5653:
@ferdikoomen Can you take a look at this PR?

qqilihq:
I'd appreciate some feedback from the maintainer whether this (or something similar to customize the HTTP client) is something which might make it into openapi-typescript-codegen. Been playing around with this yesterday and liked it quite a lot, but we need a functionality to intercept the HTTP calls.

qqilihq:
To give an update here: I was researching some alternatives to this tool in the meantime, and switched to [`oazapfts!`](https://github.com/cellular/oazapfts) which is rather simplistic, but provides a very simple hook to customize the `fetch` function. This worked great for our specific case -- maybe it's an option for the described use-cases as well.

iansan5653:
Looks like this is the way to go for now: https://github.com/ferdikoomen/openapi-typescript-codegen/issues/465#issuecomment-768278511

ferdikoomen:
@qqilihq @iansan5653 yes the comments on 456 is the way to go: https://github.com/ferdikoomen/openapi-typescript-codegen/issues/465

Issue #396, opened 2020-10-27
Optional flag which changes string to date in case of format data-time or date


NicoVogel:
In the project I am working on, we use openapi v3.
The schema contains objects which use the type `string` with the format `date-time`.

example:

```yaml
ExampleType:
  title: ExampleType
  type: object
  properties:
    id:
      type: integer
    validFrom:
      type: string
      format: date-time
      example: "2012-10-27T17:18:30.966Z"
```

We did not want to convert the date everytime and therefore we created a middleware for the server and client-side that automatically converts it from `string` to `Date`.
When creating the models via the `openapi-typescript-gencode` tool, the output models use the type `string` (as intended).
But in our case we need it to be of type `Date`, which is a special case.
Therefore an optional flag to enable this feature would be great.

**What is needed?**

It needs to remember the `format` value and in case of `date` or `date-time` the output type should be `Date` instead of `string`. 

**About implementing it**

I would try to implement it and in case that anyone wants to help me, that would be great.

NicoVogel:
I was actually able to implement it by simply passing a new flag into the handlebars and implementing the logic within the interface file ([as can be seen here](https://github.com/NicoVogel/openapi-typescript-codegen/blob/9ec2cd89aacba2451df72141b9a9e89a797d15ea/src/templates/partials/exportInterface.hbs#L13)).

Even though the code works, I am sure you would not merge such a solution. I will try to implement the logic before passing the modle to the handlebars (e.g. [here](https://github.com/NicoVogel/openapi-typescript-codegen/blob/f17161d26843314d7f3c411814398dae83314abb/src/utils/writeClientModels.ts#L22)).  But my current approach does not work yet.

NicoVogel:
Finally, I had time to implement the feature outside of hbs ([as can be seen here](https://github.com/NicoVogel/openapi-typescript-codegen/blob/option/before-hbs/src/utils/writeClientModels.ts)).
I also intent to implement some tests.

Tamno:
I would also like to have this feature. It would be interesting to merge that.

NicoVogel:
I have not had time yet to implement the tests.
Maybe next week.

ferdikoomen:
@NicoVogel Sounds interesting, i know that at my current project we are also using this formatting. However the issue i see is: How do we convert back the date to a string (when we for instance send the model back to the server). The exact date-time format is not strictly specified in the spec as far as i know, but some API's need a specific string notation:
[...]

PR #494, opened 2020-12-20
Implementation for: Optional flag which changes string to Date in case of format data-time

NicoVogel:
This is the implementation for the issue #396. 

**Functionality**

A new flag called `useDateType` is added and it is false per default. If it is used, then any generated model property which is of type `string` and format `date-time`, will be changed to `Date`.

**Usage**

This is useful if you use a response interceptor to automatically find and replace date-ISO strings with an actual Date object.

**TODO**

- [ ] Update Readme with documentation for new option and how to use

### DIFF
--- a/bin/index.js
+++ b/bin/index.js
@@ -19,6 +19,7 @@ program
     .option('--exportServices <value>', 'Write services to disk', true)
     .option('--exportModels <value>', 'Write models to disk', true)
     .option('--exportSchemas <value>', 'Write schemas to disk', false)
+    .option('--useDateType', 'Output Date instead of string for the format "date-time" in the models')
     .parse(process.argv);
 
 const OpenAPI = require(path.resolve(__dirname, '../dist/index.js'));
@@ -34,6 +35,7 @@ if (OpenAPI) {
         exportServices: JSON.parse(program.exportServices) === true,
         exportModels: JSON.parse(program.exportModels) === true,
         exportSchemas: JSON.parse(program.exportSchemas) === true,
+        useDateType: program.useDateType,
     })
         .then(() => {
             process.exit(0);
--- a/src/index.ts
+++ b/src/index.ts
@@ -24,6 +24,7 @@ export interface Options {
     exportModels?: boolean;
     exportSchemas?: boolean;
     write?: boolean;
+    useDateType?: boolean;
 }
 
 /**
@@ -52,6 +53,7 @@ export async function generate({
     exportModels = true,
     exportSchemas = false,
     write = true,
+    useDateType = false,
 }: Options): Promise<void> {
     const openApi = isString(input) ? await getOpenApiSpec(input) : input;
     const openApiVersion = getOpenApiVersion(openApi);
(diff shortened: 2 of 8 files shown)


ferdikoomen:
@NicoVogel Sorry for the late reply, so i see this an work, the only issue is serialize en deserialize flow. So i receive a date string from the backend, that gets auto converted, but then we should do the same when sending this back. That might be an issue, since date format is not standardised as far as i know...

Paden:
This use case is the only reason I decided to fork a bit back. It be great to get this incorporated. 

My client/XHR layer takes care of the back and forth @ferdikoomen describes above. Meanwhile the consumers of that client use the models (with dates where date-time formats applies).

ferdikoomen:
@Paden What do you mean with "client/XHR layer takes care of the back and forth". Did you implement a custom XHR request?

Paden:
@ferdikoomen yes, I have a custom fetch client that transforms date ISO strings to date instances. The consumer of that fetch has no idea that goes on and expects date instances for that to work properly. This is much easier than have two response types: one with dates as strings and another with dates as date instances.

cecbr:
Please, why didn't this feature go to the master branch?

cecbr:
@NicoVogel, please, why didn't this feature go to the master branch of @ferdikoomen?

xeger:
To echo something @Paden said: I only want touse `openapi-typescript-codegen` to generate models, and my client wrappers take care of Date-to-string and string-to-Date automatically. A flag such as this would allow me to use the tool without forking it.

knoefel:
I would also be very happy if this flag/option could be added! Thx in advance!

tuunit:
Hi @ferdikoomen, @NicoVogel what is holding back the merge of this issue? Can I someone fix the remaining / open points? So we could proceed with the merge?

doctau:
I also want this, but not quite the same - I want it to use the date-time types from a library not the JS standard one.

Rather than adding something specific, I think what could work is providing the ability to overridding the type mapping on https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/src/openApi/v3/parser/getMappedType.ts#L27, so the original requester can specify that `type: string; format=date-time` uses Date, but I could specify that a format maps to a library's date-time type.

CodeTherapist:
Any update? Can that be merged? @NicoVogel @ferdikoomen

rcbevans:
> I also want this, but not quite the same - I want it to use the date-time types from a library not the JS standard one.
> 
> Rather than adding something specific, I think what could work is providing the ability to overridding the type mapping on https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/src/openApi/v3/parser/getMappedType.ts#L27, so the original requester can specify that `type: string; format=date-time` uses Date, but I could specify that a format maps to a library's date-time type.

This would be really useful.

In my backend I'm already getting Dates from Prisma which get serialized, though I have to do some tricks to get them back into strings to work with the types as generated currently.

That said, on the client I use luxon rather than native js dates, so Ideally I'd like to have server side types generated which are expecting a Date, and deserialized types/clients which can get the ISO dates back into Luxon DateTimes.

Something that would allow the user to customize the mapping expected type would be great, even if to have different backend types would mean running two lots of codegen for front and backend.

jamesarosen:
@ferdikoomen I'm happy to help out here. Is there anything I can to to help get this feature shipped?

mrlubos:
@jamesarosen this feature is implemented in [our fork](https://github.com/nicolas-chaulet/openapi-typescript-codegen), can you try it and let me know if it works for you?

jamesarosen:
> can you try it and let me know if it works for you?

I've added a story to our backlog, including commenting here with results. No ETA yet.

mrlubos:
@jamesarosen no rush! Btw your website still says 2023

PR #573, opened 2021-02-18
Allow importing external references

chris-armstrong:
This is a replacement for #433 , using a library to parse external and remote $ref instances so that an openapi.yml can be split up at development time, but reassembled for the purposes of generating types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
         "commander": "^7.0.0",
         "handlebars": "^4.7.6",
         "js-yaml": "^4.0.0",
+        "json-schema-ref-parser": "^9.0.7",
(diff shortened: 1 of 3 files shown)


ferdikoomen:
Looks good @chris-armstrong

ferdikoomen:
@chris-armstrong I already bumped the version number, i was thinking should we add a small description in the README about the 'auto-bundle' feature?

chris-armstrong:
> @chris-armstrong I already bumped the version number, i was thinking should we add a small description in the README about the 'auto-bundle' feature?

Sure, I'll updated my PR with a quick note about it

PR #575, opened 2021-02-19
support for output in got request library

chris-armstrong:
Add support for outputting TypeScript code for the [`got`](https://www.npmjs.com/package/got) request library, with passthrough on the `timeout` parameter and timeout support for the other libraries


chris-armstrong:
Is this one good to go, or do you require further changes?

ferdikoomen:
@chris-armstrong The idea behind the openapi-typescript-codegen is to provide a framework agnostic client using standard fetch and xhr requests. I know there is already a slight deviation from this with the NodeJS support, but is there a good reason why we would choose **got** over the current **node-fetch** implementation? The PR itself looks fine, so that is not the problem, but maybe we can add some notes on when to select got over node-fetch?

chris-armstrong:
> @chris-armstrong The idea behind the openapi-typescript-codegen is to provide a framework agnostic client using standard fetch and xhr requests. I know there is already a slight deviation from this with the NodeJS support, but is there a good reason why we would choose **got** over the current **node-fetch** implementation? The PR itself looks fine, so that is not the problem, but maybe we can add some notes on when to select got over node-fetch?

My main motivation was to get time out and retry support, which `got` has, but I needed to add it to this PR  for the other modes. I could separate those out, and just submit the timeout and retry support separately, if that sounds reasonable?

Isn't there a way to replace the request method with a custom implementation without patching the library?

ferdikoomen:
@chris-armstrong Yes there is. You can provide a custom request function where you can use GOT / Axios / whatever. You can use this in the following way:
```
openapi-typescript-codegen --input ./spec.json --output ./dist --request ./myCustomRequest.ts
```
This will replace the current request file (that is generated in `core/request.ts`) with the file that you specified.

This is a bare bone example of a request file: https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/test/custom/request.ts
You basically receive the `ApiRequestOptions` and you need to return a `ApiResult`, anything in the middle is up to you.

chris-armstrong:
Thanks I'll close this then, hopefully the timeout support in #611 can be merged instead as it simply uses the existing providers.

PR #591, opened 2021-02-26
Wrap negative numbers in quotes, fix Enums in TypeScript

troglotit:
Negative numbers in generated enums are broken.
Here's the playground how it's not working and how PR fixes it.
https://www.typescriptlang.org/play?#code/KYOwrgtgBAouEHUCWAXAFgOUgI2AJygG8BYAKCigH0BGKAXimoBoyLKAmeqdl0gXzJlQkWPGTosEXHgBCeAPYBrUEVZUAtLQabebLY11VODHmQGkh8UZHGYc+AGJIAHsAAmq8lABElTd64dNV9qAIZmYI4w7l5zMjgbVDspfAA6GnixJMlpdPZBUgTEbPs8J1c3dOpMxIlS8vc8muK6lLKXdwBtX38AXSA

### DIFF
--- a/src/openApi/v2/parser/getEnum.ts
+++ b/src/openApi/v2/parser/getEnum.ts
@@ -11,7 +11,7 @@ export function getEnum(values?: (string | number)[]): Enum[] {
             .map(value => {
                 if (typeof value === 'number') {
                     return {
-                        name: `"_${value}"`,
+                        name: `'_${value}'`,
                         value: String(value),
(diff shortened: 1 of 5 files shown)


ferdikoomen:
@troglotit right now this means that all number enums with get quotes right? That is fine i think, but can we align on the single quote format: 
```
value: `'${value}'`
```

troglotit:
Sure. Updated

PR #592, opened 2021-02-26
- When a property has a default value it will be a non optional property

ferdikoomen:


blipk:
This is not fixed or has regressed in the latest version @ferdikoomen 

e.g.
```Python
class CompanyDetailBase(CustomBaseModel):
    CompanyID: PositiveInt


class CompanyDetail(CompanyDetailBase):
    direct_costs: Decimal = Decimal("0")
    gross_revenue: Decimal = Decimal("0")
    billable_hours: Decimal = Decimal("0")

    gross_profit: Decimal = Decimal("0")
```

Gives the following:
```Typescript
export type CompanyDetail = {
  CompanyID: number;
  direct_costs?: number;
  gross_revenue?: number;
  billable_hours?: number;
  gross_profit?: number;
};
```

Those fields should not be optional.

blipk:
Regressions are here:
https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/src/openApi/v2/parser/getModelProperties.ts#L17
https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/src/openApi/v3/parser/getModelProperties.ts#L17

It's no longer checking that  `property.default` is not undefined as it does in this PR:
https://github.com/ferdikoomen/openapi-typescript-codegen/pull/592/files#diff-c5b8aed977b3a1de23c0c6bb3121c37f37921d30831bdf537c79a200a466cbb5R18

The double negation `!!` suggest that a misconfigured eslint rule may have caused the regression e.g. https://eslint.org/docs/latest/rules/no-unneeded-ternary which will strangely correct simple ternaries to double negation. I'm not sure what might have applied to this regression though.

@ferdikoomen

PR #611, opened 2021-03-16
add timeout support

chris-armstrong:
Add basic timeout support to each of the existing client types, extracted from code in #575

### DIFF
--- a/src/templates/core/OpenAPI.hbs
+++ b/src/templates/core/OpenAPI.hbs
@@ -13,6 +13,7 @@ type Config = {
     USERNAME?: string | Resolver<string>;
     PASSWORD?: string | Resolver<string>;
     HEADERS?: Headers | Resolver<Headers>;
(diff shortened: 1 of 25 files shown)


ferdikoomen:
@chris-armstrong sorry chris, if you want a timeout on a call why don't you use the correct server-side timeout? For me this sounds like adding a feature that should be solved in a slightly different layer

chris-armstrong:
1. We use openapi-typescript-codegen on the server (to make API calls to third parties) and on the client (to call our own API)
2. It's good practice to add correctly set timeouts on server side calls as well as client side calls so that your application doesn't appear hung or hit an external timeout (e.g. a server-side request handling timeout) instead of timing out on the downstream API call.

If you still don't want this in openapi-typescript-codegen, we'll look at other options such as a custom fetch handler (which wasn't an option when I wrote this) or a vendored build.

PR #612, opened 2021-03-17
feat: new method naming

ItzNeviKat:
Method names that are now is ugly. I propose to name the method as it called in specs. In short, before:
```
UsersService.usersControllerGetInfo
```
after:
```
UsersService.getInfo
```

### DIFF
--- a/src/openApi/v3/parser/getOperationName.ts
+++ b/src/openApi/v3/parser/getOperationName.ts
@@ -6,6 +6,9 @@ import camelCase from 'camelcase';
  * the most popular Javascript and Typescript writing style.
  */
 export function getOperationName(value: string): string {
+    const parts = value.split('.');
+    if (parts.length === 2) value = parts[1];
+    
     const clean = value
         .replace(/^[^a-zA-Z]+/g, '')
         .replace(/[^\w\-]+/g, '-')


NovaStream2030:
Thanks to ItzNevikat for fixing it, personally it always pissed me off

ferdikoomen:
@ItzNeviKat @NovaStream2030 the names are taken from the open api spec itself. Do you have an example where this is generated "ugly" and did you check how this is generated in the java generator?

ItzNeviKat:
> @ItzNeviKat @NovaStream2030 the names are taken from the open api spec itself. Do you have an example where this is generated "ugly" and did you check how this is generated in the java generator? 

This is ugly, because it is already clear that the methods are called from the UsersService, but current names repeating word "users" and adds word "controller". Why "userController" word needed? It makes the name longer.

P. S. Sorry for my bad English.

ferdikoomen:
@ItzNeviKat We don't ADD the "Controller" name, that is done by the OpenAPI spec, the service name (UserService) is generated from the `tag` and the method name (usersControllerGetInfo) is taken from the `operationId`. If you don't like that name, then change this in the OpenAPI spec...

Issue #477, opened 2020-12-10
Fetch request is always sent with Accept: application/json header


iansan5653:
**Describe the bug**
I have an API that has some JSON endpoints and some file  endpoints (media type `application/octet-stream`). The generated code works for the JSON endpoints, but for the file endpoints the server returns `406` because the code always sends the request with header `Accept: application/json`.

If there is no `200` response described with media type `application/json`, the generated code should use the first media type described or it should be a method parameter.

I'm using the Fetch API with version `0.7.0`.

iansan5653:
I was able to work around this by setting the `HEADERS` config to `{"Accept": ""}`.

ferdikoomen:
@iansan5653 good point, will fix this in the next mayor release

scipper:
@iansan5653 this work around is a bit risky, because the `HEADERS` are changed for *every* request, until you reset it. So if you have multiple requests running in parallel, this work around will cause errors!

iansan5653:
@scipper I have not noticed any issues with this workaround - I just update the `HEADERS` property one single time at app startup and it works.

scipper:
@iansan5653 because an empty Accept header, as you configured it, will most likely be treated as `Accept: */*`. I did not think about that. So, as a work around, valid. 

Thanks for the hint with the `--request` parameter. I am not sure if this helps me for the moment, but I will give it a try.

ferdikoomen:
@scipper There is a new version that allows to create client instances to each client can have its own bearer token. You can specify the `--name` argument while generating the client. More info:
https://github.com/ferdikoomen/openapi-typescript-codegen#generate-client-instance-with---name-option

paladdins:
Hi! Thank you for the great tool!

I have the following schema for the endpoint that responses with  `application/octet-stream`.
```
  /pdf:
    get:
      produces:
        - application/octet-stream
      operationId: getPDF
      responses:
        200:
              schema:
                  type: string
                  format: binary
                headers:
                  Content-Disposition:
                    type: string
                    default: "attachment; filename=file.pdf"
        403:
          $ref: '#/responses/GeneralError'
        404:
          $ref: '#/responses/GeneralError'
        default:
          $ref: '#/responses/GeneralError'
```

And since functions in DefaultService are not extendable with custom headers (and I'm sure they shouldn't) - I would like it to have additional `Accept: application/octet-stream` header. wdyt about it?

ikilinc:
[...]

PR #644, opened 2021-04-07
Fix: Respect mediatype

vanhoofmaarten:
## Issue
The current version of the `openapi-typescript-codegen` does not always respect the [Media Type](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#mediaTypeObject) of the requestBody in the `Content-Type` Request header.

Example:

If we have an operation with the following `requestBody` property ...
```json
{
  "requestBody": {
    "content": {
      "application/json": {
        "schema": {
          "type": "string"
        }
      }
    }
  }
}
```

... and we use the following method in our generated client ...
```js
await api.operationMethod({ requestBody: "test" });
```

... the client will send a request with the `Content-Type` header `text/plain` because of the following code:
1. https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/src/templates/core/fetch/getHeaders.hbs#L25-L26
1. https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/src/templates/core/xhr/getHeaders.hbs#L25-L26
1. https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/src/templates/core/node/getHeaders.hbs#L25-L26

Also, the value given in the `requestBody` parameter should be passed to the `JSON.stringify` function.

## Solution
1. Parse the Media Type from the OpenAPI document.
2. Check if it is a supported media-type, like in the `getContent` function.
3. Add the media type value to a new property in the request methods `bodyMediaType`
4. Check in the core `request` functionality if `bodyMediaType` property is defined and act accordingly.
5. If not, fallback to original flow.

### DIFF
--- a/src/client/interfaces/OperationParameter.d.ts
+++ b/src/client/interfaces/OperationParameter.d.ts
@@ -3,4 +3,5 @@ import type { Model } from './Model';
 export interface OperationParameter extends Model {
     in: 'path' | 'query' | 'header' | 'formData' | 'body' | 'cookie';
     prop: string;
+    mediaType: string | null;
 }
--- a/src/openApi/v2/parser/getOperationParameter.ts
+++ b/src/openApi/v2/parser/getOperationParameter.ts
@@ -42,6 +42,7 @@ export function getOperationParameter(openApi: OpenApi, parameter: OpenApiParame
         enum: [],
         enums: [],
         properties: [],
+        mediaType: null,
     };
 
     if (parameter.$ref) {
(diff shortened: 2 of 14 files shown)


vanhoofmaarten:
I believe this is also related to #477

ferdikoomen:
@vanhoofmaarten Sorry for the late reply, I was away due to a covid infection. Thanks for the PR, looks good! I will do some quick tests and if it looks good, then will merge it today

vanhoofmaarten:
Hi @ferdikoomen, hope you're having or have had a good recovery. Thank you merging.

ferdikoomen:
@vanhoofmaarten thanks, will push a new version to NPM soon, need to check a few more tickets and PR's.

PR #645, opened 2021-04-08
Adding oneOf with discriminator property support

sjoerdmulder:
First of all awesome project! 

Currently it seems there is no support for discriminators from the V3 spec  in the generator.

This PR adds some preliminary support so that instead of:
```
export type ModelCircle = {
    kind: string;
    radius?: number;
}
export type ModelSquare = {
    kind: string;
    sideLength?: number;
}
export type Shape = Circle | Square;
```
It generates (not the the `kind: string` is replaced)
```
export type ModelCircle = {
    kind: 'circle';
    radius?: number;
}
export type ModelSquare = {
    kind: 'square';
    sideLength?: number;
}
export type Shape = Circle | Square;
```

Added some tests as well, let me know what you think!

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
         "@types/node": "15.12.1",
         "@types/node-fetch": "2.5.10",
         "@types/qs": "6.9.6",
-        "@typescript-eslint/eslint-plugin": "4.25.0",
+        "@typescript-eslint/eslint-plugin": "4.26.0",
         "@typescript-eslint/parser": "4.26.0",
         "codecov": "3.8.2",
         "eslint": "7.28.0",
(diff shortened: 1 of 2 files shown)


ferdikoomen:
@sjoerdmulder Sorry for the late reply, I was away due to a covid infection. Thanks for the PR, looks good! I will do some quick tests and if it looks good, then will merge it today

sjoerdmulder:
I hope you are feeling better!

FYI the code that the PR generates is already used by our company
internally in production with around a dozen different polymorphic types
and works without problem ;-)

Op di 27 apr. 2021 13:05 schreef Ferdi Koomen ***@***.***>:

PR #683, opened 2021-05-06
GH-681: Handle empty responses (status 204)

### LINKED ISSUE #681: Handle empty responses (status 204)
Empty response bodies (status 204) always run into an exception which is annoying while debugging, a status check should be added to `getResponseBody`, e.g.

```ts
async function getResponseBody(response: Response): Promise<any> {
    if (response.status === 204) {
        return null;
    }
    try {
        const contentType = response.headers.get("Content-Type");
        if (contentType) {
            const isJSON = contentType.toLowerCase().startsWith("application/json");
            if (isJSON) {
                return await response.json();
            } else {
                return await response.text();
            }
        }
    } catch (error) {
        console.error(error);
    }
    return null;
}
```

dbo:
fixes #681

### DIFF
--- a/src/templates/core/fetch/getResponseBody.hbs
+++ b/src/templates/core/fetch/getResponseBody.hbs
@@ -1,16 +1,18 @@
 async function getResponseBody(response: Response): Promise<any> {
-    try {
-        const contentType = response.headers.get('Content-Type');
-        if (contentType) {
-            const isJSON = contentType.toLowerCase().startsWith('application/json');
-            if (isJSON) {
-                return await response.json();
-            } else {
-                return await response.text();
+    if (response.status !== 204) {
+        try {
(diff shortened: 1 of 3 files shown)


dbo:
@ferdikoomen what can I do to drive this further?

EddyVinck:
@ferdikoomen Can this be merged? Would be nice to have support for `204`

ferdikoomen:
Sorry for the delay all. First had corona, then broke my collarbone, then we moved houses. Promise some improvements now :-)

mrlubos:
Hello @dbo @EddyVinck, I opened a pull request adding multiple types for cases where an endpoint might return both 200 and 204. Do you mind giving it a quick look if it looks correct to you? Thanks! #1656

PR #713, opened 2021-05-25
Fix: Handle nested objects and arrays in query string

omgmashina:
Implemented correct handling of nested objects and arrays in getQueryString method.
Example:
Parameter in openapi docs
```yaml
in: query
name: pagination
schema:
  type: object
  properties:
    count:
      type: integer
    offset:
      type: integer
```
Object that I pass
```json
{
  "pagination": {
    "count": 10,
    "offset": 0
  }
}
```
Before:
```?pagination=%5Bobject+Object%5D```
After:
```?pagination%5Bcount%5D=10&pagination%5Boffset%5D=0```

### DIFF
--- a/src/templates/core/functions/getQueryString.hbs
+++ b/src/templates/core/functions/getQueryString.hbs
@@ -1,19 +1,22 @@
 function getQueryString(params: Record<string, any>): string {
-    const qs: string[] = [];
-    Object.keys(params).forEach(key => {
-        const value = params[key];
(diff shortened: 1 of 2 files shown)


ferdikoomen (approved):

ferdikoomen:
@omgmashina I pushed a fix to develop that contains your idea. I moved to URLSearchParams to make the whole encoding a bit nicer to read.

PR #727, opened 2021-06-01
fix #636: condition of isRequired

### LINKED ISSUE #636: boolean property with default: false became required
My schema is:
```yaml
components:
  schemas:
    AuthRequest:
      type: object
      properties:
        email:
          type: string
          format: email
        password:
          type: string
          format: byte
          description: base64-encoded password
        remember:
          type: boolean
          default: false
      required:
        - email
        - password
```

it's generated to:
```typescript
$AuthRequest = {
    properties: {
        email: {
            type: 'string',
            isRequired: true,
            format: 'email',
        },
        password: {
            type: 'string',
            isRequired: true,
            format: 'byte',
        },
        remember: {
            type: 'boolean',
            isRequired: true,
        },
    },
};
```

`remember` is marked as `isRequired: true` but it isn't required.

tc-imba:
In #636 there is an issue about the isRequired flag.

In common sense, if a parameter has a default value, it means that this parameter is optional (not required). But in the current code, 
"isRequired: definition.default !== undefined" means that if default is not defines, it's required, making it quite strange. I think the condition should  be "definition.default === undefined" in these places.

### DIFF
--- a/src/openApi/v2/parser/getModel.ts
+++ b/src/openApi/v2/parser/getModel.ts
@@ -22,7 +22,7 @@ export function getModel(openApi: OpenApi, definition: OpenApiSchema, isDefiniti
         isDefinition,
         isReadOnly: definition.readOnly === true,
         isNullable: definition['x-nullable'] === true,
-        isRequired: definition.default !== undefined,
+        isRequired: definition.default === undefined,
         format: definition.format,
         maximum: definition.maximum,
         exclusiveMaximum: definition.exclusiveMaximum,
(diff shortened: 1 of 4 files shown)


ferdikoomen:
@tc-imba When there is a default parameter, in typescript this cannot be an optional type:

**Invalid:**
```
public static someServiceCall(parameter?: string = 'default') {
    // ...
}
```
_Error: TS1015: Parameter cannot have question mark and initializer._

**Valid:**
```
public static someServiceCall(parameter: string = 'default') {
    // ...
}
```
This is why when there is a default value (`!== undefined`) this is a required and not an optional (`?`) prop.

PR #740, opened 2021-06-12
Allow FormData as body argument

sjoerdmulder:
This PR adds support for sending a `FormData` directly as `requestBody`. Often used for file upload.

Note the warning here:  https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects that you should  not explicitly set the `Content-Type` header on the request.


EddyVinck:
This would be great to have. I'm currently running into this issue as well with a file upload.

knoefel:
+1 
Would be really great, if this could be merged as i'm facing the same issue right now. Thx!!

sjoerdmulder:
@ferdikoomen Updated this one as well!

ferdikoomen (approved):

PR #741, opened 2021-06-12
Adding oneOf with discriminator property support

sjoerdmulder:
Continuation of #645 but in a cleaned branch...

Currently it seems there is no support for discriminators from the V3 spec  in the generator.

This PR adds some preliminary support so that instead of:
```
export type ModelCircle = {
    kind: string;
    radius?: number;
}
export type ModelSquare = {
    kind: string;
    sideLength?: number;
}
export type Shape = Circle | Square;
```
It generates (not the the `kind: string` is replaced)
```
export type ModelCircle = {
    kind: 'circle';
    radius?: number;
}
export type ModelSquare = {
    kind: 'square';
    sideLength?: number;
}
export type Shape = Circle | Square;
```

Added some tests as well, let me know what you think!

### DIFF
--- a/src/openApi/v2/parser/getOperationParameterName.ts
+++ b/src/openApi/v2/parser/getOperationParameterName.ts
@@ -1,7 +1,6 @@
 import camelCase from 'camelcase';
 
-const reservedWords =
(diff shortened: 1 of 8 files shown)


sjoerdmulder:
@ferdikoomen I have updated the branch to be based on latest master

Issue #791, opened 2021-07-15
Not decoded URI component on ref


todesstoss:
Hello, trying to generate TS on completely valid OpenAPI schema having this error:

```Error: Could not find reference: "#/paths/~1admin~1cache~1clear~1%7BcacheName%7D/parameters/0"```

Looks like this ref url in not decoded properly

todesstoss:
if it could help, a similar issue:

https://github.com/drwpow/openapi-typescript/issues/587

todesstoss:
The issue is in this file:

https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/src/openApi/v3/parser/getRef.ts

path is not decoded properly.

Maybe Pointer.parse from `json-schema-ref-parser` library should be used here?

rbruggem:
Does https://github.com/todesstoss/openapi-typescript-codegen/pull/1 fix the issue?

ferdikoomen:
Hi all the fix has been merged

PR #793, opened 2021-07-16
Fix getRef on encoded  path, issue: https://github.com/ferdikoomen/op…

todesstoss:
…enapi-typescript-codegen/issues/791


rbruggem:
@ferdikoomen could this be merged if it solves this issue?
I am encountering #791.

PR #828, opened 2021-09-14
Add named types for services methods params when using '--useOptions'

motorin:
It'll be nice to have named typings for services methods params, to use them in IDE for preparing params objects in your code in advance. So, I did these changes

Prev:

``
ClientService.saveClient( {id: string, title: string} : {id?: string, title?: string} ) { ... }
```

Now:

```
export type ClientServicePostParams = {
  id?: string,
  title?: string,
}
ClientService.saveClient( {id: string, title: string}: ClientServicePostParams ) { ... }
```

### CODE REVIEW

**Comment by ferdikoomen** [comment]
@motorin In Typescript there is a very easy way to extract the "type" in this case.

```
type ClientServicePostParams = Parameters<typeof ClientService.saveClient>;
```

https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype

I would suggest using this approach, or export named 'request' models for you requests?

motorin:
@ferdikoomen — good point, thx.

But it would be nice to see this construction in generated code, rather than writing these types by hand in every project for every service method in every place, where do you want to use this.

Sure, I can put it in my own general types file (or something like this), but models and services are exactly kind of general types for a project. So it's still good idea in my mind to place this exported type near the service.

PR #850, opened 2021-10-05
fix: accept relative schema ref paths

akmjenkins:
Cool library, thanks!

I've been having trouble with $refs on my local filesystem. The tool expects the files to be relative to process.cwd(). This PR leverages json-schema-ref-parser's ability to resolve everything when given a path to a yaml or json file. 

I've chosen to preserve the existing errors, but I'm not sure they're terribly valuable because those errors don't take into account any parsing errors when the refs are resolved. I think it's possible the entire `getOpenApiSpec` function is just `RefParser.bundle` 🤔 

Maybe there's something I'm missing in these tests. Either way, this library is really awesome, I'd like to get it working with relative $refs.

### DIFF
--- a/src/utils/getOpenApiSpec.spec.ts
+++ b/src/utils/getOpenApiSpec.spec.ts
@@ -1,27 +1,52 @@
+import RefParser from 'json-schema-ref-parser';
+
 import { exists, readFile } from './fileSystem';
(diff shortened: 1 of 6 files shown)


ferdikoomen:
Closing this PR, since we now use the RefParser directly!

PR #878, opened 2021-11-08
fix(node): import missing Blob from node-fetch in node <= 15

listiani13:
Hey, thank you for this amazing package! But after upgrading to the latest version, I ran into this error:
```
 ReferenceError: Blob is not defined
[2021-11-08T15:06:31.311Z]     at isBlob (/usr/src/dist/core/request.js:55:29)
[2021-11-08T15:06:31.311Z]     at getRequestBody (/usr/src/dist/core/request.js:161:44)
```

Looks like we don't import `Blob` whilst using `isBlob` in node template.

This PR will import missing `Blob` to fix this issue. Note that we might need to move `Blob` import to `fetch-blob` later when this package support node-fetch v3

### DIFF
--- a/src/templates/core/node/request.hbs
+++ b/src/templates/core/node/request.hbs
@@ -2,7 +2,7 @@
 
 import { AbortController } from 'abort-controller';
 import FormData from 'form-data';
-import fetch, { BodyInit, Headers, RequestInit, Response } from 'node-fetch';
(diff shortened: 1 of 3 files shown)


listiani13:
Ok this wont do, `node-fetch` doesnt seem to expose their `Blob`

Issue #886, opened 2021-11-15
Allow custom service naming/formatting


bardkalbakk:
Service name generation generates redundant names when processing Open API spec generated by NSwag. NSwag will create the operationId based on `${controller}_${method}` naming, in order to ensure unique operationId's in the document.

Example Open API 3 spec:
```
 "/api/Notifications/active": {
      "get": {
        "tags": [
          "Notifications"
        ],
        "operationId": "Notifications_GetActiveNotifications",
        "responses": {
          "200": {
            "description": "",
            "content": {
              "application/json": {
                "schema": {
                  "type": "array",
                  "items": {
                    "$ref": "#/components/schemas/Notification"
                  }
                }
              }
            }
          }
        }
      }
    },
```

Will produce the following Service class by openapi-typescript-codegen:
```
export class NotificationsService {

    /**
     * @returns Notification 
     * @throws ApiError
     */
    public static notificationsGetActiveNotifications(): CancelablePromise<Array<Notification>> {
        return __request({
            method: 'GET',
            path: `/api/Notifications/active`,
        });
    }

}
```

I would like to reduce this redunant naming so that notificationsGetActiveNotifications() method is instead named 
GetActiveNotifications().

I'm not sure this can be solved in a generic way, as this `controller_method` operationId naming is just the way NSwag does things, it's not in Open API spec.

If we can have a means of doing custom naming then this can be solved on a per project basis. And that would be awsome.

[]Default NSwag operationId name generator:
https://github.com/RicoSuter/NSwag/blob/002641eb859ebd71ab69435a3f387a762b482abc/src/NSwag.CodeGeneration/OperationNameGenerators/MultipleClientsFromOperationIdOperationNameGenerator.cs#L70

ferdikoomen:
@bardkalbakk you are fixing an issue in the spec with the generator.. I get this is an 'easy' place to fix this, but it's really an issue with the generated spec. Can you inject some 'x-operation-name' keys in there? otherwise we could have a look at those. The PR you have now changes it for everybody, but some might need the name in there, for instance:

`UserService.userInfo()` or `OrdersService.ordersByName()`, we cannot assume we can strip out the "User" or "Orders" part form the operations.

bardkalbakk:
Well in this case we also control the API and spec generation with NSwag, so it was handled there*

[...]

PR #887, opened 2021-11-15
Optimize redundant operationName naming

bardkalbakk:
Simple implementation example of issue #886. Currently just implementing in OpenApi V3 parser.

Possible adjustments:
- Add runtime flag (--nswag-operationids) for doing this mutation of method names
- Check if `openApi["x-generator"].startsWith('NSwag')`  before running mutation, as an extra layer of precaution.

**Original service class without this fix:**
```
export class NotificationsService {

    /**
     * @returns Notification 
     * @throws ApiError
     */
    public static notificationsGetActiveNotifications(): CancelablePromise<Array<Notification>> {
        return __request({
            method: 'GET',
            path: `/api/Notifications/active`,
        });
    }

}
```

**With this fix:**
```
export class NotificationsService {

    /**
     * @returns Notification 
     * @throws ApiError
     */
    public static getActiveNotifications(): CancelablePromise<Array<Notification>> {
        return __request({
            method: 'GET',
            path: `/api/Notifications/active`,
        });
    }

}
```

### DIFF
--- a/src/openApi/v3/parser/getOperation.ts
+++ b/src/openApi/v3/parser/getOperation.ts
@@ -26,7 +26,13 @@ export function getOperation(
 ): Operation {
     const serviceName = getServiceName(tag);
     const operationNameFallback = `${method}${serviceName}`;
-    const operationName = getOperationName(op.operationId || operationNameFallback);
+    
+        // Remove possible redundant serviceName from operationId before generating operationName
+    let operationId = op.operationId;
+    if (operationId?.startsWith(serviceName + '_')) {
+        operationId = operationId.split(serviceName + '_')[1];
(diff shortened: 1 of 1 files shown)


ferdikoomen:
@bardkalbakk you are fixing an issue in the spec with the generator.. I get this is an 'easy' place to fix this, but it's really an issue with the generated spec. Can you inject some 'x-operation-name' keys in there? otherwise we could have a look at those. The PR you have now changes it for everybody, but some might need the name in there, for instance:

UserService.userInfo() or OrdersService.ordersByName(), we cannot assume we can strip out the "User" or "Orders" part form the operations.

PR #893, opened 2021-11-30
Allow optional request options

cmc-sreid:
I would like to be able to be able to add additional options to ApiRequestOptions which would be set on the http client when making a request. Therefore i'm submitting this PR for your review. thanks

This is to support setting additional options such as the httpsAgent if using axios. For example : -

`OpenAPI.OTHER_OPTIONS = {
    httpsAgent: new https.Agent({ rejectUnauthorized: false }),
  };`

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,18 @@ const getToken = async () => {
 OpenAPI.TOKEN = getToken;
 ```
 
+### Additional Http Request parameters
+Sometimes it might be useful to pass additional options to the HttpClient request. For example here we are setting the httpsAgent parameter on the request.
+
+```typescript
+import { OpenAPI } from './generated';
+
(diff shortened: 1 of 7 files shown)


ferdikoomen:
@cmc-sreid this only works if you have knowledge on the internals that are use. Plus the options are different for fetch, node-fetch, xhr and axios, this is a bit of a bandaid solution to a problem. What are you trying to solve here and why do you need the httpsAgent? In your case the PR: https://github.com/ferdikoomen/openapi-typescript-codegen/pull/822 would be better since you can specify your own client there

cmc-sreid:
@ferdikoomen , agreed. I'll wait on https://github.com/ferdikoomen/openapi-typescript-codegen/pull/822 being merged and use that instead. thanks

PR #929, opened 2022-01-10
Handle empty tags

keithbro:
Before this change, if your OpenAPI spec had an empty `tags` property, the service file would not be created.

After this change, the behaviour in this scenario is the same as if there was no tags property at all.

One thing to note (and I'm happy to revert it if you don't think its better), is that I've changed the default tag from "Service" to "Unnamed". The thinking behind this is that `UnnamedService` is a better name to use than `ServiceService`.

### DIFF
--- a/src/openApi/v3/parser/getServices.spec.ts
+++ b/src/openApi/v3/parser/getServices.spec.ts
@@ -1,0 +1,18 @@
+import { getServices } from './getServices';
+
+describe('getServices', () => {
+    it('should create a unnamed service if tags are empty', () => {
+        const services = getServices({
+            openapi: '3',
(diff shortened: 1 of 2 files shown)


ferdikoomen:
@keithbro I did cleanup your example and also copied it to the V2 branch of the parser. I removed the default name entirely, so it will just generate `Service` for this. You can change this name by using the `postfix` option in the generator command line

keithbro:
> @keithbro I did cleanup your example and also copied it to the V2 branch of the parser. I removed the default name entirely, so it will just generate `Service` for this. You can change this name by using the `postfix` option in the generator command line

awesome, thank you

PR #931, opened 2022-01-14
Fix: Escape template literal special characters in schema output

jurgenbelien:
If someone would use backticks in their description, it would break the generated `$<Schema>.ts` files. Escaping backticks and expression placeholders would alleviate that.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -29,6 +29,7 @@ const handlebarsPlugin = () => ({
                 knownHelpersOnly: true,
                 knownHelpers: {
                     equals: true,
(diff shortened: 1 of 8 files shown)


ferdikoomen:
@jurgenbelien I pushed a fix to the development branch based on your PR that has some more test cases and will generate correct comments and description (with their correct escaping based on the usage). Hope you will like it.

jurgenbelien:
> @jurgenbelien I pushed a fix to the development branch based on your PR that has some more test cases and will generate correct comments and description (with their correct escaping based on the usage). Hope you will like it.

Thanks!

PR #935, opened 2022-01-19
Extension options for generated service classes

kutlugsahin:
Hey Ferdi! I hope you are doing well!
First I'm very glad to see that this super handy library keeps growing and improving 🚀  Keep up the awesome work!
I'm willing to use this in my new project but it needs a little bit tuning though. So what's not enough already? Well for angular I need to have these generated service classes a little bit different. Btw it works as is also with angular but you need to compromise on angular's http service and dependency injection mechanism which makes angular devs unhappy 😭 . I also appreciate the framework agnostic structure of the current project so I have added few extension points to the exportService template where the user can inject additional code and finally an injectable angular service using http client is possible.

More details see the open issue #936

### DIFF
--- a/bin/index.js
+++ b/bin/index.js
@@ -28,21 +28,21 @@ const params = program
 const OpenAPI = require(path.resolve(__dirname, '../dist/index.js'));
 
 const genereteParams = {
-    input: program.input,
-    output: program.output,
-    httpClient: program.client,
-    useOptions: program.useOptions,
-    useUnionTypes: program.useUnionTypes,
-    exportCore: JSON.parse(program.exportCore) === true,
-    exportServices: JSON.parse(program.exportServices) === true,
-    exportModels: JSON.parse(program.exportModels) === true,
-    exportSchemas: JSON.parse(program.exportSchemas) === true,
+    input: params.input,
+    output: params.output,
+    httpClient: params.client,
+    useOptions: params.useOptions,
+    useUnionTypes: params.useUnionTypes,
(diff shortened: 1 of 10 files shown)


ferdikoomen:
@kutlugsahin Does it make sense to add an Angular client? I pushed a WIP (definitly not working yet) branch: https://github.com/ferdikoomen/openapi-typescript-codegen/tree/feature/angular

When you run `yarn run build && yarn run run` then you will see some generated services inside `/geneated/v3`

Example:
```
import { Injectable } from '@angular/core';
import { HttpClient } from '@angular/common/http';
import { Observable } from 'rxjs';

import { request as __request } from '../core/request';
import { OpenAPI } from '../core/OpenAPI';

@Injectable()
export class SimpleService {
    private readonly http: HttpClient;

    constructor(http: HttpClient) {
        this.http = http;
    }

    /**
     * @throws ApiError
     */
    public getCallWithoutParametersAndResponse(): Observable<void> {
        return __request(this.http, {
            method: 'GET',
            path: `/api/v${OpenAPI.VERSION}/simple`,
        });
    }
}
```
I'm writing the request method to work with the `Observable` pattern, then it should just be plug-and-play

ferdikoomen:
@kutlugsahin I pushed a new update in the branch, the client is WIP but should work. It supports all the same options as other clients, but is using the Angular HTTP client.

ferdikoomen:
@kutlugsahin I merged the initial version of Angular support. Im testing it further now to see how / if it works well. Closing this PR and will publish a new version to NPM soon

PR #944, opened 2022-01-26
Feature/angular

ferdikoomen:
Creating the PR for the Angular client (WIP) needs some changes on the error handling.
See https://github.com/ferdikoomen/openapi-typescript-codegen/pull/935#issuecomment-1022319303 for some discussions on the work so far


ferdikoomen:
@kutlugsahin 

The request method now looks like (added comments to make it clear, but those are not in the output):

```
/**
 * Request method
 * @param config The OpenAPI configuration object
 * @param http The Angular HTTP client
 * @param options The request options from the service
 * @returns Observable<T>
 * @throws ApiError
 */
export const request = <T>(config: OpenAPIConfig, http: HttpClient, options: ApiRequestOptions): Observable<T> => {
    const url = getUrl(config, options);
    const formData = getFormData(options);
    const body = getRequestBody(options);

    // getHeaders has been converted to observable
    return getHeaders(config, options).pipe(
        mergeMap( headers => {
            // When have received the headers, then we can send the request
            return sendRequest<T>(config, options, http, url, formData, body, headers);
        }),
        // We map the response (if successful)
        map(response => {
            const responseBody = getResponseBody(response);
            const responseHeader = getResponseHeader(response, options.responseHeader);
            return {
                url,
                ok: response.ok,
                status: response.status,
                statusText: response.statusText,
                body: responseHeader ?? responseBody,
            } as ApiResult;
        }),
        // If there is a HTTP error, then we capture it
        // If this is a error without a status (network error) we just throw it
        // Otherwise we convert it to a result object for further processing (error code could be associated with a custom error message)
        catchError((error: HttpErrorResponse) => {
            if (!error.status) {
                return throwError(() => error);
            }
            return of({
                url,
                ok: error.ok,
                status: error.status,
                statusText: error.statusText,
                body: error.statusText,
            } as ApiResult);
        }),
        // We catch the ApiResult object and check it for errors
        // If catchErrorCodes does not find anything then we return the result
        map(result => {
            catchErrorCodes(options, result);
            return result.body as T;
        }),
        // If catchErrorCodes did throw an error, then we just throw it to the subscriber
        catchError((error: ApiError) => {
            return throwError(() => error);
        }),
    );
};
```

ferdikoomen:
The client passes the test cases and you can still do things like `catchError()` and `retry()` on the observer

ferdikoomen:
note to self: need to switch to `switchMap` instead of `mergeMap`

PR #965, opened 2022-02-09
Fixes a bug that prevented using complex dictionaries within compositions

### LINKED ISSUE #964: Using complex dictionaries within compositions (oneOf, anyOf, allOf) generates wrong code
There is an issue generating code that contains complex [dictionaries](https://swagger.io/docs/specification/data-models/dictionaries/) (dictionaries that themselves contain `oneOf`, `anyOf` or `allOf`) within compositions (`oneOf`, `anyOf` or `allOf`).

This spec:
```yaml
CompositionWithOneOfAndDictionary:
  description: This is a model that contains a dictionary within composition
  type: object
  properties:
    propA:
      oneOf:
      - type: boolean
      - type: object
        additionalProperties:
          type: array
          items:
            oneOf:
            - type: boolean
            - type: string
```
should generate :
```typescript
export type CompositionWithOneOfAndDictionary = {
    propA?: (boolean | Record<string, Array<(boolean | string)>>);
};
```
but instead generates:
```typescript
export type CompositionWithOneOfAndDictionary = {
    propA?: boolean;
};
```

If i replace nested array of `boolean` or `string` into a simple array (eg. of `string`), it works as expected.

Proposed fix: #965

nikopavlica:
Fixes #964.

In case of "simple" dictionary (`Record<string, number>` or `Record<string,array<number>>`, model.type was set to `number` in `parser/getModel.ts` function. But in case of composite, `type` was  not explicitly set and defaulted to `any`.

I didn't want to change this type in the `getModel` function, as it might have some other implications, so I just added another check in when to ignore a model within composition.

Added 3 more test objects to the suite, so it covers all three cases.

### DIFF
--- a/src/openApi/v3/parser/getModelComposition.ts
+++ b/src/openApi/v3/parser/getModelComposition.ts
@@ -31,8 +31,9 @@ export const getModelComposition = (
             const hasProperties = model.properties.length;
             const hasEnums = model.enums.length;
             const isObject = model.type === 'any';
+            const isDictionary = model.export === 'dictionary';
             const isEmpty = isObject && !hasProperties && !hasEnums;
-            return !isEmpty;
+            return !isEmpty || isDictionary;
         })
         .forEach(model => {
             composition.imports.push(...model.imports);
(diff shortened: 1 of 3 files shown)


ferdikoomen:
@nikopavlica Nice work, will test the PR in some API's and let you know how it goes

vyobukhov:
@ferdikoomen any news about merging this one or maybe any help needed? 

Really waiting for it to be merged, thx!

ferdikoomen (approved):

ferdikoomen:
@vyobukhov @nikopavlica merged, will release a new version soon (0.20.0), fixing some other bugs as well

PR #980, opened 2022-03-08
return undefined explicitly to avoid compilation errors

yuvalt:


ferdikoomen:
@yuvalt where and how are you getting compilation errors? Do you have a sample project? We have numerous people using this without issues. Plus `return` or `return undefined` are really one and the same. Could it be something in your eslint or tsconfig?

yuvalt:
I get the following error:
```
sdks/axios-sdk/src/core/request.ts:129:5 - error TS7030: Not all code paths return a value.

129     return;
        ~~~~~~~
```

This is because in my `tsconfig.json` I have 

```
"noImplicitReturns": true,
```

yuvalt:
Ping @ferdikoomen

Issue #427, opened 2020-11-16
Improve build tool


yordis:
Would love to leverage the internal code for generating the TypeScript types for an alternative code-gen tool: https://github.com/straw-hat-team/openapi-web-sdk-generator. You have done an amazing job generating the TypeScript types and took some decisions around some metadata.

Fundamentally both tools do the same, different solutions when it comes to generating clients. I tried to leverage your code so since it would be much better to collaborate than re-invent the wheel.

But it seems that you compile this package into a single file, so I can't import the parser and some of your internal functions.

### Proposal

I would propose two solutions:

1. Change the build step to output multiple files rather than one file
2. Change the repo to mono-repo and create multiple packages

ferdikoomen:
@yordis the goals of this library is a fast and zero config OpenAPI client generation. This is largely done by pre-processing all Handlebar templates and making sure the dependency load in NodeJS is kept to a minimum. This is why there is just one file with a bunch of jiberish :-) but on large projects this generation speed is a huge timesaver (specially if you generate stuff during a build step). The goal is not to make a 'toolkit' or some kind of über configurable setup. For these scenarios the OpenAPI JAVA codegen is much better. 
I can imagine that we could split up the project into mono-repos and export independent packages like parser and what not.. but what would be the goal? I mean feel free to just grab what you need from my code and use it in your library :-)

ferdikoomen:
Btw you can use the library inside NodeJS like any other. The Node API also gives you the option to just generate specific parts

,,,
const path = require('path');
const openapi = require('openapi-typescript-codegen');

const spec = {
  // your spec
};

openapi.generate({
  input: spec,
  output: path.resolve(__dirname, 'generated'),
  exportCore: false,
  exportServices: true,
  // etc...
});

You can generate just the models or the services, but write your own request layer (core)

Not sure if that would work for you?

,,,

yordis:
> For these scenarios the OpenAPI JAVA codegen is much better.

But it is Java 😅 

> we could split up the project into mono-repos and export independent packages like parser and what not.. but what would be the goal?

@ferdikoomen some of my generations deviate from this one, my goal is pretty much the same, just that we have two different routes to generate the clients, folders, and what have you.

Here are some core things that the package does differently from others

- Operations are organized per file and each operation is a function (tree-shaking friendly)
- We use `x-directories` to organize the operation files
[...]

PR #992, opened 2022-03-25
Preserve modules and declarations in build & distribution

webpro:
First of all, excellent library, it was a pleasure to work with the parsed meta data from `parse()`. Thanks!

For the services output, I wanted to generate only functions (instead of classes with methods). This is efficient in terms of tree-shakeability of the resulting functions (and hooks). We previously had a solution in which we post-process the output files of this library, but by using `parse()` and the returned meta data directly, we can use our own templates to generate services and hooks very efficiently.

I've read #427. Still, I'd love to propose this PR that results in a non-bundled `dist` folder and the TS declaration files.

As it stands, I believe it's basically negligible in terms of additional costs of maintenance or performance. I've ran all tests and also used this on my own (fairly large) project without issues.

Obviously when people use internal modules from this package things may break when file names and/or structure would change (but personally I'd have no issues with this). As a follow-up you might want to consider exposing (only) the `parse` function(s), allowing to use custom templates with the results. I think this may result in folks using this library in new and interesting ways, without sacrificing the original spirit and intent.

Thanks for considering this, feel free to close it as you see fit. Then I'll likely fork and publish this version if you don't mind.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     },
     "files": [
         "bin/index.js",
-        "dist/index.js",
+        "dist",
         "types/index.d.ts"
     ],
     "scripts": {
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "scripts": {
         "clean": "rimraf ./dist ./test/generated ./test/e2e/generated ./samples/generated ./coverage ./node_modules/.cache",
         "build": "rollup --config --environment NODE_ENV:development",
+        "postbuild": "rsync -am --include='*.d.ts' --include='*/' --exclude='*' ./src/ ./dist",
         "build:watch": "rollup --config --environment NODE_ENV:development --watch",
         "release": "rollup --config --environment NODE_ENV:production",
         "validate": "tsc --project tsconfig.json --noEmit",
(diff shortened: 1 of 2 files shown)


webpro:
When `npm install`-ing this from my fork I found the `*.d.ts` source files were not included in the rollup output. Only the generated ones from the `declaration: true` setting were included.

The last commit adds them. In a rather unconventional way, as I couldn't figure out a way to do this properly using rollup, a plugin, and/or TypeScript configuration. Maybe you or someone would know of a better way to do this.

webpro:
So I figured out source DTS files are never transferred by `tsc`. They can be copied manually, or renamed from `*.d.ts` to `.ts` to include them.

As I've progressed in using only the parser and further cleaned up my clone, I've created https://github.com/webpro/parse-openapi. I've mentioned your name and repository, and added a (MIT) license.

One thing I did in my repo is to start using https://github.com/kogosoftwarellc/open-api. This seems like a solid option so you would not need to maintain them yourself.

I'd be happy to discuss further, but feel free to close this PR. I'm grateful for the inspiration that led to the other repo.

ferdikoomen:
@webpro I'm not against exposing separate files, however the API for each method is not what i consider "public" only the main method and CLI commands are supported. I know that you will understand that, but this might not be true for all others. The original ticket was about properly splitting the codebase into section, i love that idea, but right now i have no time to make a design (and do the work) for this. But i'm happy to have a look at a 'cheap' solution like you are proposing, with some warning in the README files

Issue #981, opened 2022-03-10
Response as a content "application/pdf" gets returned as response.text()


Bringordie:
Same issue as in https://github.com/ferdikoomen/openapi-typescript-codegen/issues/802 

My goal is to get a "application/pdf" and serve it to my users something like:
```javascript
const fileURL = URL.createObjectURL(data);
window.open(fileURL, '_blank');
```

I'm receiving "application/pdf" from our backend. I do not have enough knowledge of what my buddy is doing is the correct way. But he insists on sending it this way with this response header.
```json
        "responses" : {
          "200" : {
            "description" : "Successful",
            "content" : {
              "application/pdf" : {
                "schema" : {
                  "type" : "string",
                  "format" : "binary"
                }
              }
            }
          },
```

Adding something like:
```javascript
else if (contentType === 'application/pdf') {
    return await response.blob();
}
```

to getResponseBody would fix this issue:
```typescript
async function getResponseBody(response: Response): Promise<any> {
    try {
        const contentType = response.headers.get('Content-Type');
        if (contentType) {
            const isJSON = contentType.toLowerCase().startsWith('application/json');
            if (isJSON) {
                return await response.json();
            } else if (contentType === 'application/pdf') {
                return await response.blob();
            } else {
                return await response.text();
            }
        }
    } catch (error) {
        console.error(error);
    }
    return null;
}
```

I've been unable to find another way to get this to work. Googeling I've tried to make a .text() (which is what I recieve) into a Blob, but that seems to not work either.

Bringordie:
And to add extra since it only has those 2 checks. Perhaps it should be expanded to handle the rest also if this is a valid issue ? https://developer.mozilla.org/en-US/docs/Web/API/Response#methods

loganbenjamin:
I have this issue as well, I don't think that adding a check for application/pdf is the right way to do it because you'll end up needing to add a near unlimited list of mime types, plus any mime type of text/* could be text or blob depending on how the data is being used.

I think the best way would be to check for the response type (binary) and handle accordingly.
I've done that and created a PR: #986

Issue #829, opened 2021-09-16
Cannot use headers AND content in responses


ricardojosegomezulmke:
Currently generated code returns either 1 (one) response header or the body as defined in spec.
But it cannot return BOTH, headers AND body.

Example from sample spec v3:
GET /user/login
responses 200:
"responses": {
 "200": {
     "description": "successful operation",
     "headers": {
         "X-Rate-Limit": {
             "description": "calls per hour allowed by the user",
             "schema": {
                 "type": "integer",
                 "format": "int32"
             }
         },
         "X-Expires-After": {
             "description": "date in UTC when toekn expires",
             "schema": {
                 "type": "string",
                 "format": "date-time"
             }
         }
     },
     "content": {
         "application/xml": {
             "schema": {
                 "type": "string"
             }
         },
         "application/json": {
             "schema": {
                 "type": "string"
             }
         }
     }
 },

In the example, the generated user login service will only return the content. There is no mechanism to get the headers 
X-Rate-Limit or X-Expires-After.
The only way to achieve this today is to specify 3 services, e.g.:
GET /user/login => returns the content
GET /user/login/x-rate-limit: specify ONLY the header X-Rate-Limit
GET /user/login/x-expires-after: specify ONLY the header X-Expires-After

and make 3 calls.

ferdikoomen:
@ricardojosegomezulmke That is correct, this is by design for this client generator.

adriengibrat:
Why this design?

They are various case where it's useful to get body & headers.
I'm trying to find any benefits to access only one or the other, but I'm confused.

In my case, I want to handle pagination with the [link headers](https://www.rfc-editor.org/rfc/rfc5988.html#section-5) and this particular client design force me to make the same (costly) request twice!

**Are you open to a PR**?
I'm willing to provide the following changes that update this behaviour to:
-  [add a new `headers` property in `ApiResult` type](https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/src/templates/core/ApiResult.hbs#L8)
-  change accordingly all `request` implementations ([exemple](https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/src/templates/core/fetch/request.hbs#L79))
[...]

PR #1043, opened 2022-05-17
Pdf file request with response headers

bertramr:
We have a similar problems like #829 and #981, when trying to access an endpoint with a pdf file and the filename in the header.

Therefore I started working on this branch and fixed it (at least for our problem). Please be aware that this pull request at the moment would introduce breaking change and I doubt it is in the sense of the original design. So in effect I don't expect this to be merged but I still hope it might be useful for someone.

- ApiResult is now a Generic with body and header. 
- During `getOperation` both the header and the content from the responses are pushed to the list, handled to create the necessary entries and finally the header is dropped, in order to not polluting the types.
- A new request function is added, which is chosen, when a responseHeader is included in the options. With that it only includes still only one header

Known short comings:
 - Still only handles one header
 - It is always expected that the first entry is the entry with the binary type, which is not guaranteed to best of my knowledge
 - If a responseHeader is set, the return type is now always a ApiResult and no longer a string. 
 - It's missing additional test cases to verify the added behavior

This branch includes other than stated in the contribution guideline merge commits from the two open pull requests (#1037 and #986 )

Would be happy about any feedback. 

Best regards and thank you very much for this neat code generator

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## unreleased
+### Fixed
+- Fixed binary response type to process as a Blob rather than text/string (#986)
+- Use type-only imports for axios (#1037)
+
 ## [0.22.0] - 2022-04-26
 ### Fixed
 - Upgraded dependencies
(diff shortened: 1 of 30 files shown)


ufon:
nice

PR #1048, opened 2022-05-23
Fix error TS2564

huzaifahj:
Error on `tsc`:

```
openapi-client/core/CancelablePromise.ts:25:14 - error TS2564: Property '[Symbol.toStringTag]' has no initializer and is not definitely assigned in the constructor.

25     readonly [Symbol.toStringTag]: string;
```


KiwiKilian:
To add some more information, this error occurs when using TypeScript v 4.7.2 – works fine with previous versions.

ferdikoomen (approved):

PR #1067, opened 2022-06-07
Prevent duplication of parameter names

kulczy:
Hi

I am working with an API that has parameters like e.g `product.code` and `product.code[]` The problem is that when generated, these parameters are doubled. I added a simple workaround. I don't know if this is the best solution, but it works for me, so I will leave it here and I hope it helps 😄


dennispassway:
I need this as well :)

ferdikoomen:
@kulczy Thanks!

PR #1073, opened 2022-06-09
Feature/arrayParamNaming

allartk:
[The search filter of openApi platform](https://api-platform.com/docs/core/filters/#search-filter) uses a questring with both `key` for a single value, and `key[]` for filter on multiple values. Currently this generates duplicated operation params in the services. A way around is pluralizing param names with square brackets.


allartk:
I just checked the existing PR, this PR  overlaps with https://github.com/ferdikoomen/openapi-typescript-codegen/pull/1067/files

ferdikoomen:
@allartk merged https://github.com/ferdikoomen/openapi-typescript-codegen/pull/1067/files because it's a more straightforward way and does not need an additional library to support this

PR #1141, opened 2022-07-14
Add model postfix

### LINKED ISSUE #983: Adding model suffix option
Hello, great project, thanks! Have you considered adding an option for setting up the suffix for generated model classes inside `models` folder? I have a lot of conflicting names in my codebase after generating the sdk from openapi schema - e.g. my custom `Offer` class and then generated `Offer` model class. I was thinking of create my own file where I will re-export these model classes with the suffix (e.g. `OfferModel`), but it seems to my a bit unfortunate to maintain, etc. so that is why I am asking about that. Thanks for answer!

markbrockhoff:
Closes #983 

I implemented the option to add a suffix to the generated model names.
To stay aligned with the already exisiting parameter --postifx, I named it --postfixModels and renamed --postfix to --postfixServices.
The --postfix parameter is still supported so this is not a breaking change. I just marked it as deprecated in the docs and the --help command so the --postfixServices parameter should be prefered over it.

In order for this feature to be as unintrusive as possible, I implemented it by renaming the imported model inside index.ts.
This way all logik about the generated models stays the same, just the clash of model names inside index.ts can be prevented.


asabates:
@ferdikoomen would be lovely to have this merged and released 🙏

ferdikoomen:
I merged the PR, but will need to update the main branch, will push a new release soon

davneto:
Hi, this feature is not working for me for some reason. I'm allowed to include the --postfixModels parameter with corresponding string value, but the generated models are not being generated with the postfix. The latest release already includes this feature fully implemented? Any ideas what's going on? Thank!

aryobenholzner:
@davneto I experienced the same.

The problem is, that the suffix is not being applied to the type directly, but rather only to the export in the `index.ts`.

This means the `postfixModels` flag is behaving differently to `postfixServices` which indeed renames the generated classes and filename.

I think this should be fixed to bahave the same.

CC: @ferdikoomen just that you are aware of this

pneme:
Could anyone please fix this? It blocks me from using this library :(

Issue #1154, opened 2022-07-18
Support `additionalProperties: true`


ajenkins-cargometrics:
In OpenAPI 3, a schema like:

```yaml
    AnyValues:
      description: An object whose property values can be of any type
      type: object
      additionalProperties: true
```

Should be equivalent to a typescript type like:

```typescript
type AnyValues = {
    [key: string]: any
};
```

However what openapi-typescript-codegen actually generates is a type like this:

```typescript
type AnyValues = {
};
```

That is, an object with no properties.

Additionally, OpenAPI allows combining `additionalProperties: true` with explicit property definitions.  So an OpenAPI schema like this:

```yaml
SomeType:
    type: object
    required: [name]
    properties:
        name:
            type: string
    additionalProperties: true
```

Should result in a typescript type like:

```typescript
type SomeType = {
    name: string,
    [key: string]: any
};
```

but in fact it results in:

```typescript
type SomeType = {
    name: string
};
```

Desired result: `additionalProperties: true` in an object schema should result in a `[key: string]: any` property in the corresponding typescript type.

garcipat:
This is seems to be related to this one: #1066

ajenkinski:
> This is seems to be related to this one: #1066

That issue is different.  In that case, the problem is that the spec doesn't specify a type for the schemas.  That is, it has types defined like

```yaml
  schemas:
    Account:
      properties:
        balance:
          type: string
          description: balance in unit WEI, presented with hex string
          example: '0x47ff1f90327aa0f8e'
        energy:
          type: string
          description: energy in uint WEI, presented with hex string
          example: '0xcf624158d591398'
        hasCode:
          type: boolean
          description: whether the account has code
          example: false
```
Simply adding `type: object` to the type fixes the problem.  I'm not sure if the `type` field is supposed to be optional, but in any case it's a different problem.

mrlubos:
Hey @ajenkins-cargometrics, if this is still relevant, we got this feature in [our fork](https://github.com/nicolas-chaulet/openapi-typescript-codegen) that is kept in sync with this repository, would love you to give it a try

PR #1165, opened 2022-07-27
Support free-form objects

kshramt:
> A free-form object (arbitrary property/value pairs) is defined as:
>
>     type: object
>
> This is equivalent to
>
>     type: object
>     additionalProperties: true
>
> and
>
>     type: object
>     additionalProperties: {}

https://swagger.io/docs/specification/data-models/data-types/

### DIFF
--- a/test/__snapshots__/index.spec.ts.snap
+++ b/test/__snapshots__/index.spec.ts.snap
@@ -3643,6 +3643,9 @@ export { EnumWithExtensions } from './models/EnumWithExtensions';
 export { EnumWithNumbers } from './models/EnumWithNumbers';
 export { EnumWithStrings } from './models/EnumWithStrings';
 export type { File } from './models/File';
+export type { FreeFormObjectWithAdditionalPropertiesEqEmptyObject } from './models/FreeFormObjectWithAdditionalPropertiesEqEmptyObject';
+export type { FreeFormObjectWithAdditionalPropertiesEqTrue } from './models/FreeFormObjectWithAdditionalPropertiesEqTrue';
+export type { FreeFormObjectWithoutAdditionalProperties } from './models/FreeFormObjectWithoutAdditionalProperties';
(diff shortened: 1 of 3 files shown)


test/__snapshots__/index.spec.ts.snap
```diff
@@ -4334,6 +4340,42 @@ export type File = {
...
+
+exports[`v3 should generate: ./test/generated/v3/models/FreeFormObjectWithAdditionalPropertiesEqTrue.ts 1`] = `
+"/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
+/**
+ * This is a free-form object with additionalProperties: true.
+ */
+export type FreeFormObjectWithAdditionalPropertiesEqTrue = Record<string, any>;
```
rvadim, line 4363: Hi! Why are you using Record type?
AFAU Record used to fix fields in a object. https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

But additionalProperties are about free keys in object.

Thank you!

kshramt: Thank you for the comment.

> Hi! Why are you using Record type?

IIUC, the use case for `Record` is not limited to restricting keys to the union type of literals. According to https://stackoverflow.com/a/54104476, `Record<string, any>` is equivalent to `{[key: string]: any}`. Also, https://github.com/ferdikoomen/openapi-typescript-codegen/blob/2721045415f87ca971a09e5fc6976caa4b2c944f/test/spec/v3.json#L1831-L1837 is converted to a `Record` (https://github.com/ferdikoomen/openapi-typescript-codegen/blob/2721045415f87ca971a09e5fc6976caa4b2c944f/test/__snapshots__/index.spec.ts.snap#L1220).

kshramt:
https://github.com/ferdikoomen/openapi-typescript-codegen/issues/1154 might be resolved by the PR.

Issue #1182, opened 2022-08-05
Add option to use ESM style import specifiers


Awem:
The code generated by this package is currently not suitable to be easily used in ESM projects. This is because for ESM, we need different import specifiers: https://nodejs.org/api/esm.html#import-specifiers

So e.g.:

`export { ApiError } from './core/ApiError` should be `export { ApiError } from './core/ApiError.js`

It would be helpful to provide a configuration option, that changes the generated code accordingly.

Pustur:
Seems like #1226 is trying to fix this

klippx:
@ferdikoomen Can you please re-open this?

klippx:
Here is a codesandbox illustrating the problem: https://codesandbox.io/p/sandbox/openapi-typescript-codegen--with-esm-c83t3c?file=%2Fsrc%2Findex.ts%3A1%2C1

See how we cannot build since the generated types are not ESM? Hence we cannot get a dist folder updated without manually updating the generated files.

matthewadams:
@ferdikoomen We are trying to upgrade to Node 20.x & Typescript 5.x and this is a blocker.  It seems to me that there are two issues here:  first, having the tool generate code that is usable in an ESM project, and second, _using_ this tool in an ESM project.

It appears that https://github.com/ferdikoomen/openapi-typescript-codegen/pull/1226 is attempting to address the second issue.  There still needs to be a PR to address the first.

doichev-kostia:
If it's blocking for someone, you can patch the package with https://gist.github.com/doichev-kostia/ccfb0689a5c72f66950518f062249e43 where I added the `.js` extension to those imports in the bundled code

for instance:
```shell
pnpm patch openapi-typescript-codegen@<version>
```

You'll receive something like
```shell
You can now edit the following folder: /private/var/folders/nr/bj8rh6yn4yn2l16wlfjygmlw0000gn/T/1c9e286e3d40a1ea67821c5c574c7a02

Once you're done with your changes, run "pnpm patch-commit /private/var/folders/nr/bj8rh6yn4yn2l16wlfjygmlw0000gn/T/1c9e286e3d40a1ea67821c5c574c7a02"
```

[...]

PR #1226, opened 2022-08-26
Add .js extension to imports to support ES Modules

macalinao:


asafkotzer:
Hey @ferdikoomen 👋 any objections to merging this? We're also depending on this PR. Thanks!

ferdikoomen:
How are you consuming the output, inside a nodeJS application? Because the requirement to have `.js` inside imports is a Node specific issue, maybe we should limit these changes to the nodeJS client (or have it as a build flag?) Right now it throws all kinds of import warnings on older / bundled / browser based projects

bulolo:
any progress?

winterrobert:
Is there something stopping this from being merged? Blocking our adoption atm

svenheden:
I'm very interested in this as well

sam-mfb:
so are we.  we use this package to create an isomorphic api client and so being able to use node esm resolution would be nice.

klippx:
What issue are you trying to solve here? I fail to see how this will solve anything, as it is just a PR that links to an ESM page and this project is not an ESM project.

What we really want is a solution to this issue #1182

src/templates/index.hbs
```diff
@@ -5,13 +5,13 @@ export { {{{clientName}}} } from './{{{clientName}}}';
```
doichev-kostia, line 5:
```hbs
export { {{{clientName}}} } from './{{{clientName}}}.js';
```

src/templates/client.hbs
```diff
@@ -4,13 +4,13 @@
...
+import { BaseHttpRequest } from './core/BaseHttpRequest.js';
+import type { OpenAPIConfig } from './core/OpenAPI.js';
+import { OpenAPI } from './core/OpenAPI.js';
 {{else}}
-import type { BaseHttpRequest } from './core/BaseHttpRequest';
-import type { OpenAPIConfig } from './core/OpenAPI';
+import type { BaseHttpRequest } from './core/BaseHttpRequest.js';
+import type { OpenAPIConfig } from './core/OpenAPI.js';
 import { {{{httpRequest}}} } from './core/{{{httpRequest}}}';
 {{/equals}}
```
doichev-kostia, line 16:
```hbs
{{#if services}}
{{#each services}}
import { {{{name}}}{{{@root.postfix}}} } from './services/{{{name}}}{{{@root.postfix}}}';
{{/each}}
{{/if}}
```

should be
```hbs
{{#if services}}
{{#each services}}
import { {{{name}}}{{{@root.postfix}}} } from './services/{{{name}}}{{{@root.postfix}}}.js';
{{/each}}
{{/if}}
```

PR #1250, opened 2022-09-08
Set default providedIn to root for angular client

sheshnathverma:
Set default providedIn to root for angular client


sheshnathverma:
Hi @ferdikoomen, Could you please review and merge it, it is beneficial for angular application/lib

PR #1363, opened 2022-12-13
add more detail to Generic Error message for easier troubleshooting

LittleUmbrella:
'Generic Error' gives no insight into the nature of the issue, whereas having at least the http status conveys a great deal for troubleshooting


ferdikoomen:
@LittleUmbrella Thanks! I will make a small update to the code, but the idea is very usefull

PR #1389, opened 2023-01-03
Add docker wrapper

lee-elenbaas:
Add docker wrapper to be able to use it as a cli tool without direct installation on the host


ferdikoomen:
@lee-elenbaas thanks!

Issue #1412, opened 2023-01-19
error: unknown option '--input


gavinthomas-valtech:
**Describe the bug**
I have installed the package (npm install openapi-typescript-codegen --save-dev) and run the command

```npx openapi --input ./openapi.yaml --output ./generated --client axios```

I get 
```error: unknown option '--input'```

gavinthomas-valtech:
ignore me.. the docs state to use openapi but it worked for me with

```px openapi-typescript-codegen --input ```

PR #1416, opened 2023-01-24
Rename bin script

naokie:
## Description of Problem

I tried to use this with redocly/redocly-cli and had problems because the bin script is provided under the same name.

cf. https://github.com/Redocly/redocly-cli/blob/main/packages/cli/package.json#L7

## Proposed Solution

In my personal opinion...
I would prefer that the primary name of the bin script be the same as the name of the package.

## Additional Information

related #1412

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,9 @@ npm install openapi-typescript-codegen --save-dev
 ## Usage
 
 ```
-$ openapi --help
+$ openapi-typescript-codegen --help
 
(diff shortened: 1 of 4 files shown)


ferdikoomen:
@naokie There where numerous requests to change this, however given the number of implementations already out there its not possible to change this. For a next big iteration of the library this could be changed. There are already some issues on Github about this, so for now i will close this PR

PR #1425, opened 2023-01-28
Aggregate header promises

m1heng:


aloisklink:
> sorry for the late reply, but what is the advantage of using `Promise.all()` in this case?

I'm not related to @m1heng, but using [`Promise.all()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) runs all of the promises in parallel:

> > It is typically used when there are multiple related asynchronous tasks that the overall code relies on to work successfully — all of whom we want to fulfill before the code execution continues.
>
> _Taken from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#description_

Image that you had to request a TOKEN from an auth server, and you use the [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) HTTP header to cache responses when possible:

```js
import {setTimeout} from "node:timers/promises";
const confg = {
  TOKEN: async() => {
    await setTimeout(1000); // fake a slow connection to auth server
    return "my-constantly-changing-token";
  },
  HEADERS: async() => {
    await setTimeout(1000); // fake a slow connection to cache DB
    return {ETag: "my-hash-of-already-cached-data"};
  },
};
```

Currently, it will take 2000ms to get the token, then the headers, but if you use `Promise.all()`, it will only take 1000ms, since they run in parallel.

ferdikoomen:
@m1heng there is a conflict in the PR, but I get your idea, i will merge the changes manually.

Issue #1428, opened 2023-02-02
Support onDownloadProgress in ApiRequestOptions


lukqw:
**Describe the solution you'd like**
I'd like to make axios requests with the `onDownloadProgress` option. 
For that I need `ApiRequestOptions` to include `readonly onDownloadProgress?: (progressEvent: ProgressEvent) => void;`
Furthermore I need this field to be unwrapped into the `requestConfig` of the `sendRequest` method.

Maybe in general a way of extending axios request options would be nice?

Or-Zarchi-Forter:
I've created a PR that can solve this by allowing an axios custom instance, #1434. Please comment there so it will get attention 😄

50Bytes-dev:
Check my PR #1682

Issue #1084, opened 2022-06-15
Support upload/download progress handlers


jgoz:
It would be great if the `axios` and `xhr` handlers supported per-request upload/download progress handlers. This is supported in the official openapi-generator package via client-specifc options that can be passed through to each request.

I'm not sure what would be the easiest path to support these handlers in this project. Fetch, Node, and Angular don't support progress callbacks, so maybe the quickest path would be to follow openapi-generator and accept optional request overrides on each request. These could even be opt-in via CLI arguments.

E.g.,

```ts
myService.uploadFile(fileBlob, {
  onUploadProgress: (progress) => {},
  // ... other client-specific request options
})
```

fugudesign:
+1

sourcec0de:
Some cases could be handled through providing a custom axios instance since `onUploadProgress` is a constructor option. https://github.com/ferdikoomen/openapi-typescript-codegen/issues/942

Or-Zarchi-Forter:
I've created a PR that can solve this by allowing an axios custom instance, #1434. Please comment there so it will get attention 😄

50Bytes-dev:
Check my PR #1682

jfroffice:
any update ?

50Bytes-dev:
> Check my PR #1682

Just use my PR #1682

50Bytes-dev:
> any update ?

Just use my PR #1682

Issue #942, opened 2022-01-26
Support for using custom instance of axios


rchl:
**Describe the solution you'd like**
When using the `httpClient: 'axios' option, I would like to be able to also use a custom axios instance to make the requests.

The reason for that is that I would like to use an Axios instance configured to my liking (custom interceptors, different timeout value, etc...).

Looking at the code it seems that it would be pretty easy to support that and I could contribute if the feature request is accepted.

ferdikoomen:
How are you configuring Axios, globally or per instance? because you can already set global interceptors.

rchl:
I'm creating a custom Axios instance and configuring it.

smarlhens:
Hi there,

I may be interested in using a custom instance of Axios to override configuration of httpsAgent in order to get rid of the error `ERR_CERT_AUTHORITY_INVALID` when using auto-signed SSL certificates.

```typescript
const httpsAgent = new Agent({
  rejectUnauthorized: false,
});
```

Not sure this is the best way but I guess it would be nice if we could provide a custom instance.

PS: Apologize if this is not the right place to put that message, as the title is a little bit generic, I wanted to add my use case.

Wish you a nice day 🌞

busheezy:
You can specify your axios interceptors and defaults before axios is used. There isn't a way for different settings for different named clients so this won't work for some use cases.

https://share.busheezy.dev/bOEKbJ.png

IniterWorker:
I second @rchl, `axios` without a way to custom the current `axiosInstance` is not meaningful. But, I understand that the encapsulation HttpRequest doesn't want to carry custom behaviour from `axios`. So, what is the best approach to use `AxiosInstance` vs `axios` global?

AllanPinheiroDeLima:
Having an option to use custom instances would be great. I actually have a problem where a can't use interceptors nor set language headers in a nuxt app because the framework encapsulates the options from the composables. 

Unless I'm doing something very wrong, I can't have multiple configs working for all parts of my app without dropping this lib entirely

tibbe:
I ran into this as well. In addition to the general reasons for supporting a custom axios instance mentioned above (i.e. don't mutate global state that can be used by any part of the program), having a custom axios instance is required if you want to automate JWT token refresh (using interceptors) in case the current token has expired. You can't have the global instance put your credentials on all requests and not all request are hitting the same server (and thus need different credentials).

sam-mfb:
[...]

PR #1434, opened 2023-02-05
Allow usage of a custom axios instance

Or-Zarchi-Forter:
This PR solves #942, and any issues requesting specific advanced axios features (e.g. #1428, part of #1084)
 
Currently, it is impossible to add any interceptors to axios without modifying the global axios instance, which can effect other parts of your program in unwanted ways.

Also, hiding away the actual usage of the axios library in the `request` file means this generator will always play catch up with new axios features, or people asking for very specific axios parameters to be externalized.

This PR allows passing in a custom axios instance to the inner request module, providing an escape hatch for advanced users.

### DIFF
--- a/docs/axios-support.md
+++ b/docs/axios-support.md
@@ -24,3 +24,51 @@ in your `tsconfig.json` file:
     }
 }
 ```
+
+
+## Using a custom Axios client
+
+Sometime you may want to use your own Axios client created by `axios.create` for advanced configuration (e.g. Usage of the popular [axios-retry](https://github.com/softonic/axios-retry) interceptor) without having to [reimplement](./custom-request-file.md) the entire generated Axios request function.
+
+In those cases, simply construct your own HttpRequest wrapper implementation and pass it into your API client
+
+## Example
+
+Create a file that looks like this, that references file from the `/core` folder of the generated client.
+
+```typescript
+
+import axios from 'axios';
+import axiosRetry from 'axios-retry';
+import { request as __request } from './request';
+import { CancelablePromise } from './CancelablePromise';
+import { BaseHttpRequest } from './BaseHttpRequest';
+import { ApiRequestOptions } from './ApiRequestOptions';
+import type { OpenAPIConfig } from './OpenAPI';
+
+
+export class AxiosHttpRequestWithRetry extends BaseHttpRequest {
+  axiosInstance = axios.create();
+
+  constructor(config: OpenAPIConfig) {
+    super(config);
+    axiosRetry(this.axiosInstance);
+  }
+
+  public override request<T>(options: ApiRequestOptions): CancelablePromise<T> {
+    return __request(this.config, options, this.axiosInstance);
+  }
+}
+
+```
+
+Then, when instantiating your generated test client, pass in your custom request wrapper class:
+
+```typescript
+
+import { AxiosHttpRequestWithRetry } from './AxiosRequestWithRetry';
+import { GeneratedClient } from './generated/client';
+
+const client = new GeneratedClient({ BASE: 'http://localhost:8123' }, AxiosHttpRequestWithRetry)
+
+```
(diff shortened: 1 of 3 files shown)


alondayan2 (approved):

blipk:
There's already docs and instructions for something like this: https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/docs/custom-request-file.md

Looks like the dev has possibly abandoned this repo though, there are breaking bugs that havent had attention in months.

rchl:
> There's already docs and instructions for something like this: [`master`/docs/custom-request-file.md](https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/docs/custom-request-file.md?rgh-link-date=2023-02-07T00%3A17%3A16Z)

Having to re-implement whole request logic means that there is a very high chance of things breaking on new version of this package being released.

This change looks like it would be a lot less prone to that.

Although ideally there would be some axios-client-specific method that one could override like `createAxiosInstance` so that it would be even more future proof but I imagine that would need more refactoring and potentially more axios-client-specific logic.

Or-Zarchi-Forter:
> There's already docs and instructions for something like this: [master/docs/custom-request-file.md](https://github.com/ferdikoomen/openapi-typescript-codegen/blob/master/docs/custom-request-file.md)

I've mentioned this method in the docs i've added. But as @rchl mentioned, that process is very error-prone, and when attempting it I ended up copying hundreds of codes from the default generated version of `request.ts`

yehonatanz:
Would really love to this merged as well.
My use case has a very carefully configured axios client going through a application-level proxy (not a standard HTTP proxy).
The complexity of copying all the code from the default `request.ts` file made me give up and opt for a different openapi client generator instead.
It would be really nice if this got merged and I could re-consolidate my openapi clients.

blipk:
@yehonatanz 
Which API generator did you go with instead?
Curious as this hasn't had any updates in a while

yehonatanz:
> @yehonatanz Which API generator did you go with instead? Curious as this hasn't had any updates in a while

@blipk [this one](https://github.com/OpenAPITools/openapi-generator-cli) with the `typescript-axios` generator selected.
Pretty versatile, but has the distinct disadvantage of being written in Java and requiring a JRE to run (also I find it generates a more awkward interface, but that's a matter of taste I guess)

esbenam:
Would love to see this merged - as others have already mentioned, this is a blocker for some people using customised `axios` instances. Anything we can do to help this along? Is this implementation considered a viable option, or do the maintainers have a different view of how this should be implemented/handled?

Copying and overriding the request file won't do the trick in my case, since I am generating a client used by multiple projects, each of them having the need to pass in a custom `axios` instance when they generate the client (setting default custom headers etc.)

thegris:
I am also looking for this functionality, thank you!

Or-Zarchi-Forter:
@ferdikoomen any chance this can be merged in? it's a relatively simple change, and it seems like it addresses a common pain point

chazmuzz (approved):

atomicleopard:
Would be awesome to see some movement on this, such a small change but such a big impact 🙏

Or-Zarchi-Forter:
thank you @ferdikoomen for merging this PR  🎈

jm-bairesdev:
Does this work for non-named clients too (`--name` is not used), where all we have is static public functions in our generated services/clients?
We don't instantiate the client, so we don't have a way to:

`const client = new GeneratedClient({ BASE: 'http://localhost:8123' }, AxiosHttpRequestWithRetry)`

shervinw:
@ferdikoomen  I don't think that this implementation solves all cases when needing to provide a custom axios instance, since the axios instance instantiation can only occur inside the BaseHttpRequest override, it results in a too tightly coupled solution.

The axios instance would be better provided as an injected dependency in the constructor, so that we can create and pass in the instance from anywhere.

I am forking this repo to solve the problem, with backward compatibility, I'll create a PR and would be great if it could be considered to merge

PR #1442, opened 2023-02-15
[request] export used functions for reusability

edwinveldhuizen:
When using a custom request method, you might want to use some or even most of the methods generated with this client.
This PR makes sure it exports all used functions in request.ts so you can import it from a custom request

### DIFF
--- a/src/templates/core/angular/sendRequest.hbs
+++ b/src/templates/core/angular/sendRequest.hbs
@@ -1,4 +1,4 @@
-export const sendRequest = <T>(
+export export const sendRequest = <T>(
 	config: OpenAPIConfig,
 	options: ApiRequestOptions,
 	http: HttpClient,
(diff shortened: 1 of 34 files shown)


src/templates/core/fetch/sendRequest.hbs
```diff
@@ -1,4 +1,4 @@
-export const sendRequest = async (
+export export const sendRequest = async (
```
edwinveldhuizen, line 1:
```suggestion
export const sendRequest = async (
```

src/templates/core/angular/sendRequest.hbs
```diff
@@ -1,4 +1,4 @@
-export const sendRequest = <T>(
+export export const sendRequest = <T>(
```
edwinveldhuizen, line 1:
```suggestion
export const sendRequest = <T>(
```

src/templates/core/axios/sendRequest.hbs
```diff
@@ -1,4 +1,4 @@
-const sendRequest = async <T>(
+export export const sendRequest = async <T>(
```
edwinveldhuizen, line 1:
```suggestion
export const sendRequest = async <T>(
```

Issue #1448, opened 2023-03-01
CancellablePromise in Typescript 4.9.5


mstosio:
In Typescript version 4.9.5, after generting there is error in CancellablePromise:

readonly [Symbol.toStringTag]: string;

TS2564: Property '[Symbol.toStringTag]' has no initializer and is not definitely assigned in the constructor.

ferdikoomen:
This should be fixed in the latest release @mstosio

ferdikoomen:
Thanks for the PR!

PR #1449, opened 2023-03-01
Update toStringTag in CancellablePromise

mstosio:
This PR resolves problem with issue #1448.

In typescript 4.9.5 after calling generate CancellablePromise has typescript error :
S2564: Property '[Symbol.toStringTag]' has no initializer and is not definitely assigned in the constructor.


yuhr:
@ferdikoomen Could you merge this please?

PR #1468, opened 2023-04-08
Fix TanStack Query compatability by adding support for returning null instead of undefined

Shinevision:
### Issue
This pull request aims to address an issue that arises when an endpoint returns a `204 No Content` status. Currently, the TanStack query generates an error stating that the Query data cannot be undefined, as demonstrated in the following screenshot:
afbeelding

### Solution
To resolve this issue, the proposed solution involves allowing the return of null values instead of undefined by introducing an optional parameter. This change will enable the library to handle `204 No Content` responses more gracefully and prevent errors in such scenarios.

### Changes

This pull request includes the following changes to the codebase:

- Added --useNullForNoContent flag in the README, documentation, and command-line options

- Added useNullForNoContent to the Options type, the generate function, and related utility functions
- Updated test cases to include the --useNullForNoContent flag
- Modified the getResponseBody template for one of the core clients (applies to all) to conditionally return null or undefined based on the useNullForNoContent flag

These changes allow the library to support "204 No Content" responses by returning null instead of undefined when the --useNullForNoContent flag is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ $ openapi --help
     --name <value>            Custom client class name
     --useOptions              Use options instead of arguments
     --useUnionTypes           Use union types instead of enums
+    --useNullForNoContent     Use null for returning instead of undefined if a response returns 204 no content
     --exportCore <value>      Write core files to disk (default: true)
     --exportServices <value>  Write services to disk (default: true)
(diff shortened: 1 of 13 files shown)


mrlubos:
Hi @Shinevision, wondering if you've raised an issue with https://github.com/TanStack/query about this, too?

PR #1497, opened 2023-04-25
Indicate auto-generation in file headers.

gz:
Thanks for this great code generator.

I wonder if you would consider this addition, it adds a line to the file header indicating that the file is auto-generated. Feedback I got from my collaborators is that it was not obvious that files created by openapi-typescript-codegen were automatically generated.

### DIFF
--- a/src/templates/partials/header.hbs
+++ b/src/templates/partials/header.hbs
@@ -1,3 +1,4 @@
+/* generated using openapi-typescript-codegen -- do no edit */
 /* istanbul ignore file */
 /* tslint:disable */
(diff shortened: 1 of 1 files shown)


src/templates/partials/header.hbs
```diff
@@ -1,3 +1,4 @@
+/* generated using openapi-typescript-codegen -- do no edit */
```
mrlubos, line 1: Late to the party, but I see a typo in here @gz 

```
/* generated using openapi-typescript-codegen -- do not edit */
```

PR #1572, opened 2023-05-30
fix: pascal case

Mokto:
First, thanks for your library, I'm REALLY happy to be able to remove the java dependency from our codebase!

---------------

**Motivation**: the most popular golang openapi plugin, https://github.com/swaggo/swag, generates references in that format:

`packagename.myFunction`.

This library then generates types with the name: `packagename_myFunction`.

My PR is addressing that issue and would generate `PackagenameMyFunction`

-----------------

1. Any typescript interface should be written in pascal case, in my opinion.
2. The java library was already changing those types into pascal case.

Note: this change is really major and should be considered as a breaking change. I can potentially add a new parameter to the command line tool to be compatible with the previous versions.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+node_modules
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,10 @@
 FROM node:alpine
 WORKDIR /usr/src/openapi
+
+COPY ["package.json", "package-lock.json", "./"]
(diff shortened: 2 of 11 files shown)


ferdikoomen:
@Mokto We should definitely have a flag for this, since its a breaking change. Would you be able to add this? Ideally we could have a flag called `--case`, with options for `pascal`, `snake`, etc.?

Issue #1626, opened 2023-07-05
CancelablePromise fails to resolve with Next.js 13


pitmullerIngka:
**Describe the bug**
I've tried to upgrade an application from Next.js 12 (v12.3.4) to 13 (v13.4.8) and suddenly experienced issues when fetching data from the API through the generated typescript client.
It still send out the requests and I got a response from the API, but the Promise would not resolve and just stay stuck there. I did not see or receive error messages, but even after waiting minutes, there would never be a response.

Going into the generated code, it seemed to fail resolving in `CancelablePromise.ts` in the constructor trying to call the `resolve` function on the `Promise`. I troubleshooted with some console.logging to debug the issue, like the following (CancelablePromise.ts:40-50):

```typescript
this.#promise = new Promise<T>((resolve, reject) => {
      this.#resolve = resolve;
      this.#reject = reject;

      const onResolve = (value: T | PromiseLike<T>): void => {
        console.log("starting to resolve");
        if (this.#isResolved || this.#isRejected || this.#isCancelled) {
          return;
        }
        console.log("resolving");
        this.#isResolved = true;
        this.#resolve?.(value);
        console.log("resolved");
      };
```

The console logs:

```
13:57:08.464 starting to resolve [CancelablePromise.ts:45:16]
13:57:08.465 resolving [CancelablePromise.ts:49:16]
```

This issue also seems to occur for `reject`.

OS: Linux Mint 21.1 Cinnamon
Browser: Firefox 114.0.2

Tested also on a Mac with Chromium, which showed an error message on row 56:
```typescript
this.#reject?.(reason);
```
```
TypeError: (intermediate value)(intermediate value)(intermediate value) is not a function
```

Now, I have no clue why this issue occurs when updating Next.js for me, but something in the background must've changed that I'm not aware of. The exact same generated code works for me with Next.js 12.
It seems to try to call the `resolve()` function before the instance is initiated and not resolving the valid syntax of `this.#resolve?.(value)` correctly.

As said, no clue why this happens, but doing the `undefined` check without an intermediate value, seems to resolve this issue for me, so I would propose this in a PR and would be interested in your opinions on this.

So instead of `this.#resolve?.(value);`, I propose:

```typescript
if (this.#resolve) this.#resolve(value);
```

moee:
[...]

PR #1627, opened 2023-07-05
fix: adjust promise resolve for next issue

pitmullerIngka:
Adressing issue #1626.

Proposal for a slight adjustment of the resolving process for the `CancelablePromise`.

As said, I have no clue why this issue occurs when updating Next.js for me, but something in the background must've changed that I'm not aware of. The exact same generated code works for me with Next.js 12.
It seems to try to call the `resolve()` function before the instance is initiated and not resolving the valid syntax of `this.#resolve?.(value)` correctly.

Doing the `undefined` check without an intermediate value, seems to resolve this issue for me, so I would propose this in a PR and would be interested in your opinions on this.
Imo, this shouldn't affect other users in a negative way and only increase the code size very slightly.

So instead of `this.#resolve?.(value);`, I propose:
```typescript
if (this.#resolve) this.#resolve(value);
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "openapi-typescript-codegen",
-    "version": "0.24.0",
+    "version": "0.25.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "openapi-typescript-codegen",
-            "version": "0.24.0",
+            "version": "0.25.0",
             "license": "MIT",
             "dependencies": {
                 "camelcase": "^6.3.0",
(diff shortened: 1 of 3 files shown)


maggo:
There's an additional conditional call at [Line #123](https://github.com/ferdikoomen/openapi-typescript-codegen/blob/41e25b86e22a70f342a1cf42f91f957d9c2b14f5/src/templates/core/CancelablePromise.hbs#L123C8-L123C15) that needs to be fixed

mfrlic:
+1, this fixes the issue

ferdikoomen:
Thanks all!

brunouber:
Hi @ferdikoomen 
Is it possible you didn't publish a new version on npm since this fix has been merged?

Graunephar:
Waiting for a new version also :)

kristw:
+1 Would appreciate new version.

Issue #1643, opened 2023-07-11
Null types are not handled


rveciana:
I'm using FastAPI 100, that creates OpenAPI version 3.1 as an output.

I want an object with arrays with elements of a type or null, such as:

data: dict[str, list[float|None] | list[int|None]]

Before, the None was ignored when generating the OpenAPI file, but now, the generated file looks like:

```
"additionalProperties": 
              "anyOf": [
                {
                  "items": {
                    "anyOf": [{ "type": "number" }, { "type": "null" }]
                  },
                  "type": "array"
                },
                {
                  "items": {
                    "anyOf": [{ "type": "integer" }, { "type": "null" }]
                  },
                  "type": "array"
                }]}
```

Then, openapi-typescript-gen creates wrong types like this: 

    data: Record<string,> 

instead of:
    
        data: Record<string, (Array<number|null> | Array<string|null>)>;

Looking at the code, I can't see the place where openapi-typescript-gen processes the type null.

PR #1683, opened 2023-08-07
support nested any-of to use with openapi v3.1 type null

smarlhens:
Hi there 👋🏻 

Thanks @ferdikoomen for your amazing job on this library 🙏🏻 

I've tried my best to implement support for nested anyOf in order to use the[ type null rework on open api specification v3.1](https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0).

It seems related to https://github.com/ferdikoomen/openapi-typescript-codegen/issues/1643.

I remain at your disposal for any feedback on the implementation.

I wish you a great day ☀️ 

Best regards ☮️

### DIFF
--- a/src/openApi/v3/parser/getModelComposition.ts
+++ b/src/openApi/v3/parser/getModelComposition.ts
@@ -30,9 +30,10 @@ export const getModelComposition = (
         .filter(model => {
             const hasProperties = model.properties.length;
             const hasEnums = model.enums.length;
+            const hasLink = typeof model.link !== 'undefined' && model.link !== null;
             const isObject = model.type === 'any';
             const isDictionary = model.export === 'dictionary';
-            const isEmpty = isObject && !hasProperties && !hasEnums;
+            const isEmpty = isObject && !hasProperties && !hasEnums && !hasLink;
             return !isEmpty || isDictionary;
         })
         .forEach(model => {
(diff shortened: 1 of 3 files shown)


PaulVrugt:
I have a nested `anyOf` within an `oneOf`. Like:

```
"SomeDto": {
        "type": "object",
        "properties": {
          "value": {
            "oneOf": [
              {
                "type": "integer",
                "format": "int32"
              },
              {
                "type": "string"
              },
              {
                "type": "array",
                "items": {
                  "anyOf": [
                    {
                      "type": "string"
                    },
                    {
                      "type": "number"
                    }
                  ]
                }
              }
            ],
            "nullable": true
          }
        },
        "additionalProperties": false
      },
```

The generated api code seems to ignore the AnyOf, and generates the type as:

```
export type SomeDto = {
  value?: (number | string) | null;
};
```

instead of:

```
export type SomeDto = {
  value?: (number | string | Array<number | string>) | null;
};
```

does this PR also fix this issue?

smarlhens:
hey @PaulVrugt,

with this PR changes, it will generate: 
```typescript
export type SomeDto = {
    value?: (number | string | Array<(string | number)>) | null;
};
```

you can check on your machine if you want:
```bash
git clone git@github.com:smarlhens/openapi-typescript-codegen.git
git checkout feat/support-nested-any-of
npm install
```

then add at the end of `components`.`schemas` inside `test/spec/v3.json`:
```json
"SomeDto": {
    "type": "object",
    "properties": {
        "value": {
            "oneOf": [
                {
                    "type": "integer",
                    "format": "int32"
                },
                {
                    "type": "string"
                },
                {
                    "type": "array",
                    "items": {
                        "anyOf": [
                            {
                                "type": "string"
                            },
                            {
                                "type": "number"
                            }
                        ]
                    }
                }
            ],
            "nullable": true
        }
    }
}
```

and then run
```bash
npm run test:update
git diff
```

you will see that inside `test/__snapshots__/index.spec.ts.snap`, in line to be added, the export type of `SomeDto` will be equal to the first snippet I wrote.

PaulVrugt:
Hi @smarlhens ,

so the short answer is `yes` right? :)

PaulVrugt:
@ferdikoomen can we get this merged? We are running into this in our code

smarlhens:
> Hi @smarlhens ,
> 
> so the short answer is `yes` right? :)

Yes 👍🏻

PR #1865, opened 2023-11-05
feat(operationalParameter): Default to config headers for header params

optikalefx:
If OpenAPI has paramaters that will be sent as headers, allow configuration to set those headers for the entire client.

This PR generates the following typescript

```ts
 public listWorkflows({
        instanceId,
        amazonAdvertisingApiClientId = this.httpRequest.config.HEADERS['Amazon-Advertising-API-ClientId'],
        amazonAdvertisingApiAdvertiserId = this.httpRequest.config.HEADERS['Amazon-Advertising-API-AdvertiserId'],
        amazonAdvertisingApiMarketplaceId = this.httpRequest.config.HEADERS['Amazon-Advertising-API-MarketplaceId'],
        nextToken,
        limit,
    }
```

Which allows a user to use the library as follows

```ts
const myClient = new MyClient({
		TOKEN: token,
		HEADERS: {
			'Amazon-Advertising-API-ClientId': AMAZON_ADS_CLIENT_ID,
			'Amazon-Advertising-API-AdvertiserId': AMAZON_ADS_ADVERTISING_ID,
			'Amazon-Advertising-API-MarketplaceId': AMAZON_ADS_US_MARKETPLACE
		}
	});
```

And then method calls no longer require you to pass these headers every time. This is a welcome change to verbosity.

Feedback is welcome.

### DIFF
--- a/src/openApi/v3/parser/getOperationParameter.ts
+++ b/src/openApi/v3/parser/getOperationParameter.ts
@@ -54,7 +54,11 @@ export const getOperationParameter = (openApi: OpenApi, parameter: OpenApiParame
             operationParameter.base = model.base;
             operationParameter.template = model.template;
             operationParameter.imports.push(...model.imports);
-            operationParameter.default = getModelDefault(schema);
+            operationParameter.default =
+                parameter.in === 'header'
+                    ? `this.httpRequest.config.HEADERS['${parameter.name}']`
+                    : getModelDefault(schema);
+
             return operationParameter;
         } else {
             const model = getModel(openApi, schema);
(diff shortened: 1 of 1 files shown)


optikalefx:
To the maintainers, there are at least 3 ways I could think of to implement this.

1) What you see in this PR, which is simply overriding the default in `getOperationParameter.ts`. This is the least code changed, but doesn't make the most sense overall.

2) Making the change in `getModelDefault.ts`. Which makes more sense, but requires us to have access to `parameter` not schema, in order to access `in`. I did try this change first, to pass `parameter` instead of `definition: OpenApiSchema`. However this requires changing getModel and `getOperationParameter` in quite a number of places.

3) Making a new partial (this was my first approach) called `{{>defaultParameters}}` and doing this work there in the hbs. However, we don't love logic in the hbs, and {{{default}}} already handles the case where `?:` has to be applied vs just `:` So it was a bit of repeat logic that default already implements.

At any rate, I went with the least changes necessary for now, but happy to discuss the other approaches. I could make 3 PRs to look at if you want to, but before I write a test, I wanted to check with you on approach.

Thanks!

Issue #1884, opened 2023-11-10
Option to convert constant size arrays into tuples


orhanhenrik:
When generating types for a constant size array, I would like to have an option to convert it to a constant size array / tuple in typescript.

## Example
```ts
MyModel: {
  type: 'object',
  properties: {
    coordinates: {
      type: 'array',
      items: {
        type: 'number',
      },
      minItems: 2,
      maxItems: 2,
    },
  },
},
```
Would convert to 
```ts
export type MyModel = {
  coordinates: Array<number>;
}
```
currently, but I would like the option to convert it to
```ts
export type MyModel = {
  coordinates: [number, number];
}
```

PR #1885, opened 2023-11-10
Add support for converting constant size arrays into tuples

orhanhenrik:
Fixes https://github.com/ferdikoomen/openapi-typescript-codegen/issues/1884

This PR implements a new option `--useTuples` which converts constant size arrays into constant size typescript arrays.

I'm not too happy with how the logic ended up partially in handlebars and partially in typescript, so if you have any suggestions for how I could make this work better with the project's code structure I would appreciate that! The problem I was facing was that I didn't have access to any options inside `getModel` without having to pass the `useTuples` option throughout the whole program.

## Example
```ts
MyModel: {
  type: 'object',
  properties: {
    coordinates: {
      type: 'array',
      items: {
        type: 'number',
      },
      minItems: 2,
      maxItems: 2,
    },
  },
},
```
Without `--useTuples`
```ts
export type MyModel = {
  coordinates: Array<number>;
}
```
With `--useTuples`
```ts
export type MyModel = {
  coordinates: [number, number];
}
```


nicholaschiang:
+1

orhanhenrik:
Works in the maintained fork of this library https://github.com/hey-api/openapi-ts

PR #1913, opened 2023-11-28
fix: support chinese for enum

wleven:
support chinese string for enum


wleven:
@ferdikoomen review

src/openApi/v2/parser/getEnum.ts
```diff
@@ -20,7 +20,7 @@ export const getEnum = (values?: (string | number)[]): Enum[] => {
                 }
                 return {
                     name: String(value)
-                        .replace(/\W+/g, '_')
+                        .replace(/[^a-zA-Z0-9_\u4e00-\u9fa5]+/g, '_')
```
PetrusAsikainen, line 23: If the name is used as an [identifier](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#identifiers), the correct characters to include would be the ones with the ID_Start or ID_Continue Unicode property. The CJK characters in the BMP [with ID_Start](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5Cp%7BID_Start%7D) are currently `\u3400-\u4DBF\u4E00-\u9FFF\uF900-\uFAD9`.

josstn:
I had a similar need to support nordic characters, so based on @PetrusAsikainen's mdn link about identifiers above, I've created another PR that I believe handles all valid identifier characters: #2038 

Maybe that could be merged instead of this one?

PR #1268, opened 2022-09-22
Add option to pass in a custom handlebars template to generate the service files

mb21:
This is a rough draft PR with what we're currently using, if you're interested in merging this, can be refined.

We're currently using this with this template:

```hbs
{{!-- this file is based on https://github.com/ferdikoomen/openapi-typescript-codegen/blob/75a648cc4e2e90cf78b26fcfe8b641797d82ad86/src/templates/exportService.hbs --}}
/* tslint:disable */
/* eslint-disable */
{{#if imports}}
{{#each imports}}
import type { {{{this}}} } from '../models/{{{this}}}';
{{/each}}

{{/if}}
import { Result } from '@/definitions/BasicTypes';
import { Opts } from '@/helpers/http/http'
import { openApiRequest } from '@/helpers/http/openApiRequest';

{{#each operations}}
/**
{{#if deprecated}}
  * @deprecated
{{/if}}
{{#if summary}}
  * {{{escapeComment summary}}}
{{/if}}
{{#if description}}
  * {{{escapeComment description}}}
{{/if}}
{{#unless @root.useOptions}}
{{#if parameters}}
{{#each parameters}}
  * @param {{{name}}} {{#if description}}{{{escapeComment description}}}{{/if}}
{{/each}}
{{/if}}
{{/unless}}
{{#each results}}
  * @returns {{{type}}} {{#if description}}{{{escapeComment description}}}{{/if}}
{{/each}}
  */
const {{{name}}} = ({{>parameters}}fetchOptions?: Opts
): Promise<Result<{{>result}}>> => {
  return openApiRequest({
    method: '{{{method}}}',
    url: '{{{path}}}',
    {{#if parametersPath}}
    path: {
      {{#each parametersPath}}
      '{{{prop}}}': {{{name}}},
      {{/each}}
    },
    {{/if}}
    {{#if parametersCookie}}
    cookies: {
      {{#each parametersCookie}}
      '{{{prop}}}': {{{name}}},
      {{/each}}
    },
    {{/if}}
    {{#if parametersHeader}}
    headers: {
      {{#each parametersHeader}}
      '{{{prop}}}': {{{name}}},
      {{/each}}
    },
    {{/if}}
    {{#if parametersQuery}}
    query: {
      {{#each parametersQuery}}
      '{{{prop}}}': {{{name}}},
      {{/each}}
    },
    {{/if}}
    {{#if parametersForm}}
    formData: {
      {{#each parametersForm}}
      '{{{prop}}}': {{{name}}},
[...]

PR #1995, opened 2024-01-23
Add option to pass in custom handlebars templates #1268

gersongoulart:
Introducing a simple, yet dynamic solution to conquer issue #1268 with unparalleled flexibility! 🚀🌟

# Overriding Built-In Templates
The parameter `templateOverrides` allows the override of *any* template or partial with a custom implementation.

### package.json
```json
{
    "scripts": {
        "generate": "openapi --input ./spec.json --output ./generated --templateOverrides index:\"Hello World\" service:./templates/service.hbs"
    },
}
```

### NodeJs
```typescript
const OpenAPI = require('openapi-typescript-codegen');

OpenAPI.generate({
    input: './spec.json',
    output: './generated',
    templateOverrides: {
        index: 'Hello World', // the template itslef
        service: 'templates/service.hbs', // or a file path
    },
});
```

### DIFF
--- a/src/utils/preCompileTemplate.ts
+++ b/src/utils/preCompileTemplate.ts
@@ -1,0 +1,35 @@
+import { readFileSync } from 'fs';
+import handlebars from 'handlebars';
+import { extname } from 'path';
+
+/**
+ * Precompiles a Handlebars template from a given file.
+ *
+ * @param {string | undefined} file - The path to the file containing the Handlebars template.
+ * @returns {TemplateSpecification | undefined} - The precompiled template as a string, ready to be exported.
+ */
+export function preCompileTemplate(file?: string): TemplateSpecification | undefined {
+    if (!file) return;
+
+    const template = extname(file) === '.hbs' ? readFileSync(file, 'utf8').toString().trim() : file;
+    const templateSpec = handlebars.precompile(template, {
+        strict: true,
+        noEscape: true,
+        preventIndent: true,
+        knownHelpersOnly: true,
+        knownHelpers: {
+            ifdef: true,
+            equals: true,
+            notEquals: true,
+            containsSpaces: true,
+            union: true,
+            intersection: true,
+            enumerator: true,
+            escapeComment: true,
+            escapeDescription: true,
+            camelCase: true,
+        },
+    });
+
+    return eval(`(function(){return ${templateSpec} }());`);
+}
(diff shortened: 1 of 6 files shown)


src/utils/preCompileTemplate.ts
```diff
@@ -0,0 +1,35 @@
...
+            union: true,
+            intersection: true,
+            enumerator: true,
+            escapeComment: true,
+            escapeDescription: true,
+            camelCase: true,
+        },
+    });
+
+    return eval(`(function(){return ${templateSpec} }());`);
```
sudhir4ev, line 34: `eval` seems safe enough, although I suspect code quality tools will complain.

Here is an another _(extremely hacky)_ alternative using a temp file which gets resolved 

```ts
import { readFileSync, writeFileSync, mkdtempSync } from 'fs';

...

    const tempDir = mkdtempSync('template-');
    const tempFilePath = join(tempDir, 'template.js');
    writeFileSync(tempFilePath, `module.exports = ${compiled};`);

    const module = require(resolve(tempFilePath));
    execSync(`rm -rf ${tempDir}`);
    return module;
```

src/utils/registerHandlebarTemplates.ts
```diff
@@ -83,8 +83,101 @@ import partialTypeInterface from '../templates/partials/typeInterface.hbs';
 import partialTypeIntersection from '../templates/partials/typeIntersection.hbs';
 import partialTypeReference from '../templates/partials/typeReference.hbs';
 import partialTypeUnion from '../templates/partials/typeUnion.hbs';
+import { preCompileTemplate } from './preCompileTemplate';
 import { registerHandlebarHelpers } from './registerHandlebarHelpers';
 
+export type TemplateOverrideNames =
```
sudhir4ev, line 89: Perhaps this type can be added to the `Options` type for `generate` api in `types/index.d.ts`

src/utils/preCompileTemplate.ts
gersongoulart: Thanks for the observation. Yeah, code quality tools will warn about the safety of this to prevent developers from shipping code that could be exploited. Seeing this is a library and the provider of the template to be evaluated is the developer using the library, I think this is a safe implementation of the eval function — and the cleanest solution I could find. Writing and reading to disk on every precompilation would significantly affect performance and might encounter problems with disk access permissions.

sudhir4ev: Same thoughts. I wish we could use some other api besides `handlebars.precompile`. But I could not find any that works.

salzhrani:
can we assist with getting this PR reviewed and merged?

sudhir4ev:
Any update on this? Bump!

ng-state:
any updates?

mrlubos:
@ng-state this package is no longer maintained, see README

PR #2031, opened 2024-02-08
Fix: pass `isSuccess` to axios call

rossmacarthur:


mrlubos:
@rossmacarthur which issue does this fix?

rossmacarthur:
@mrlubos Without this you can't really change the behaviour of `isSuccess` because what axios considers an error response diverges from the `request.ts` code. It makes much more sense to use the same function for determining success and the axios signature is the same.

mrlubos:
@rossmacarthur any chance you can share an example or add a test this fixes? I'd be happy to merge this into [our fork](https://github.com/nicolas-chaulet/openapi-typescript-codegen)

jordanshatford:
To anyone still looking for this, check out our (previously forked) version of this package https://github.com/hey-api/openapi-ts. We have just recently added support for setting request/response interceptors. In the case of axios, this would allow you to modify the config used for each request before the request is sent. Please check it out if you'd like, and create any issues if you run into problems.

PR #2043, opened 2024-02-19
allow operationId be ignored when generating operation names

askvortcov:
Adds a CLI parameter `--ignoreOperationId`,  which allows ignoring OperationId in the swagger and relies on the path to generate operation names.

Use path placeholders to decorate operation names.

Ignoring operation ID is handy when working with a swagger generated by `springdoc-openapi`, since it generates messy operation IDs when confronted with multiple possibilities of `Accept` header.

### DIFF
--- a/bin/index.js
+++ b/bin/index.js
@@ -19,6 +19,7 @@ const params = program
     .option('--exportCore <value>', 'Write core files to disk', true)
     .option('--exportServices <value>', 'Write services to disk', true)
     .option('--exportModels <value>', 'Write models to disk', true)
+    .option('--ignoreOperationId <value>', 'Use operation id', true)
     .option('--exportSchemas <value>', 'Write schemas to disk', false)
     .option('--indent <value>', 'Indentation options [4, 2, tabs]', '4')
     .option('--postfixServices <value>', 'Service name postfix', 'Service')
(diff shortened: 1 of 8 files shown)


bin/index.js
```diff
@@ -19,6 +19,7 @@ const params = program
     .option('--exportCore <value>', 'Write core files to disk', true)
     .option('--exportServices <value>', 'Write services to disk', true)
     .option('--exportModels <value>', 'Write models to disk', true)
+    .option('--ignoreOperationId <value>', 'Use operation id', true)
```
mrlubos, line 22:
```suggestion
    .option('--ignoreOperationId <value>', 'Ignore operation id', true)
```

Although I agree with the initial intent – make the flag opt-in and default to true to preserve current functionality

```suggestion
    .option('--useOperationId <value>', 'Use operation id', true)
```

mrlubos:
@askvortcov want to open this pull request against [our fork](https://github.com/nicolas-chaulet/openapi-typescript-codegen)?

bin/index.js
askvortcov: Fixed in the next commit. Thanks for the review.

askvortcov:
> @askvortcov want to open this pull request against [our fork](https://github.com/nicolas-chaulet/openapi-typescript-codegen)?

Done. There is a merge conflict though. If I merge it it pulls too many commits along.

mrlubos:
@askvortcov no worries, I will resolve it!

Issue #2063, opened 2024-03-13
Update sendRequest for axios to align on parameter deprecation (withCredentials)


evgarthub:
Hi @ferdikoomen!

Axios fixed vulnerability related to `withCredentials` request parameter by introducing new parameter `withXSRFToken`, details here:
https://github.com/axios/axios/pull/6046
https://github.com/advisories/GHSA-wf5p-g6vw-rhxx

Does it make sense to include new parameter next to `withCredentials` as author suggests?

We verified internally that indeed passing both parameters with true value prevents client from behaving differently after axios update.
![image](https://github.com/ferdikoomen/openapi-typescript-codegen/assets/23404448/20087809-e7d7-4ff7-808e-73850aa68781)

I will quickly create PR for it, let me know if it does make sense?

mrlubos:
@evgarthub I think we'd want a separate parameter for this. It's a breaking change in Axios, therefore it should be a breaking change in this package, too.

evgarthub:
@mrlubos do you mean we should extend config with `WITH_XSRF_TOKEN`?

mrlubos:
Sounds like it. I'd need to dig more into it, but if this feature is only available in Axios, config should also have that flag only when using Axios client. That's what it looks like to me right now

evgarthub:
I'm not able to run any tests, or rather they are failing because of #2052 issue (I'm on Windows).

jordanshatford:
@evgarthub The windows issue should be fixed in our fork https://github.com/hey-api/openapi-ts. We have also added support for interceptors, so if you would like to do something like this you can see examples in this ticket:

https://github.com/hey-api/openapi-ts/issues/76

evgarthub:
Hi @jordanshatford ! Thanks for offering a solution! But it's already been fixed in this repo.

mrlubos:
@evgarthub yes, except it isn't actively maintained and still has many other bugs that are fixed in openapi-ts https://github.com/ferdikoomen/openapi-typescript-codegen/issues/2064

Issue #2052, opened 2024-03-01
Wrong path on Windows


mmospanenko:
On Windows, when use app under another process, RefParser gets wrong path. And as result - it joins abs and rel paths from package and argument. So I had to fix it like this:
<img width="734" alt="image" src="https://github.com/ferdikoomen/openapi-typescript-codegen/assets/2766211/60394ee9-7e52-47dc-9650-6f942e615242">

STR:
1. create file with exec (cli or import - the same)
2. run as node myfile.js on windows

<img width="1849" alt="image" src="https://github.com/ferdikoomen/openapi-typescript-codegen/assets/2766211/56bf0232-2d94-4105-9491-f5db1d76d760">

mrlubos:
Thanks. Seems quite a few people run into this issue

mrlubos:
@tajnymag hm, this package used that version before and people still complained so it got reverted

tajnymag:
> @tajnymag hm, this package used that version before and people still complained so it got reverted

Interesting. I'll try to find out why it worked and why it would't. I tried it with openapi-codegen (your version) installed as a devDependency

mrlubos:
@tajnymag thanks, let me know if you find out please. I didn't get around to trying this one out yet

mrlubos:
@tajnymag i suspect we'll need to fork the other package too https://github.com/APIDevTools/json-schema-ref-parser/issues/303

tajnymag:
@mrlubos I think I fixed the issue. Basically followed @mmospanenko's suggestion and made all passed paths to json-schema-ref-parser absolute.

Both of these had to be made together to pass our tests
1. always pass absolute url to json-schema-ref-parser
2. update json-schema-ref-parser to v11.1.0

mrlubos:
@tajnymag is the absolute path conversion happening inside this package or is that something you do before passing the value to `--input`?

mrlubos:
@mmospanenko we've published a fix in [our fork](https://github.com/nicolas-chaulet/openapi-typescript-codegen), mind giving it a try and see if it works without the need for workarounds on your end?

mmospanenko:
I can't test on Windows, it was funny testing game while we were trying to solve the issue)) but yes, will be great to remove workaround, thank you, @mrlubos!
I think thread could be closed if you've fixed this

mrlubos:
[...]

PR #2053, opened 2024-03-02
Fixed issue 2052

tajnymag:
Both of these had to be made together to pass our tests
1. always pass absolute url to json-schema-ref-parser
2. update json-schema-ref-parser to v11.1.0

Should fix: https://github.com/ferdikoomen/openapi-typescript-codegen/issues/2052

PR #2067, opened 2024-03-14
fix: updated axios sendRequest (#2063)

evgarthub:
Fix for #2063 
added backwards compatibility for axios vulnerable parameter

I checked all the changes for axios, and it looks like setting it to true when credentials are forced for any domain is the right choice. 

I'm not able to run the tests properly as #2052 causes wrong paths to be resolved.
I will take a look at that PR and see if it helps.


evgarthub:
Not sure how to force build, it has timeout on `npm install`. 
I've cherry picked #2053 and run the tests - all passed locally.

mrlubos:
Which commands did you run @evgarthub?

evgarthub:
I followed `CONTRIBUTING.md`, npm run test after build is done

PR #2095, opened 2024-03-27
TEC-XXXX: add exportSchemaArray, which puts all the schemas into an array

joshformationgames:
- We need to be able to feed a list of all of our schemas into Z-Schema for validation of uploaded JSON files
- We want this process to happen automatically for new/modified/deleted items
- Therefore, add exportSchemaArray, which exports all the schemas into a list that can be consumed by Z-Schema


joshformationgames:
Apologies, accidentally PR'd against this rather than forked version

PR #2111, opened 2024-04-15
Appen support Blob type

vicimpa:


mrlubos:
Hey @vicimpa this repository won't be actively maintained but it has a successor at https://github.com/hey-api/openapi-ts. If you're interested in this feature and we don't have it yet, please open an issue there

vicimpa:
I will try. This week we just need to decide what to do next.

PR #2184, opened 2024-06-09
opposite

brunokrebs:


mrlubos:
@brunokrebs would you like to provide context?

brunokrebs:
I'm sorry, @mrlubos, I thought it would just run CI to show an issue with that. The test I updated was the wrong one anyway.

So, being less lazy (sorry about the laziness before), I clone into my computer, added a test case, and generated the snapshot that shows what I was going to report (and possibly fix?).

As you can see, if we have an enum with keys like `+name` and `-name`, and `+total` and `-total`, the library ends up ignoring the "opposite" key (of whatever comes first, + or -).

so, on the snapshot, we ended up with:

```ts
export enum EnumWithStringsWithPlusAndMinusSigns {
    _NAME = '+name',
    _TOTAL = '+total',
}
```

instead of a type with four keys. Well, after looking closer at the generated enum/type (happens with unions as well, which is what I'm using), I realized that the issue might not be that straightforward to fix and that I should do something on my end? Anyway, again, gonna take a closer look.

mrlubos:
@brunokrebs you got it with the deprecation notice. Feel free to open an issue in Hey API if this isn't fixed there yet.

PR #2250, opened 2024-09-25
fix: changed exportService to destructure DTOs

martinshhh:


martinshhh:
changed exportService to destructure DTOs

PR #2275, opened 2024-12-23
Update org name

deekshith1856:
update org name in package.json


mrlubos:
@deekshith1856 check out the README file if you're interested in an actively maintained project

PR #2808, opened 2026-09-04
Fix code injection via unescaped spec-derived strings in generated clients

carfeii:
See the linked issue for the vulnerability report and impact.

## Root cause

Several templates interpolate OpenAPI-spec-derived strings directly into a single-quoted JS string literal in the generated output, with no escaping: the request path (`url: '{{{path}}}'` in `exportService.hbs`), each path/query/header/cookie parameter's wire name, the request-body media type, the response header name, and the first server URL / API version (`client.hbs`, `core/OpenAPI.hbs`). None of the corresponding parser values (`getServices.ts`'s `url`, `getServer.ts`'s return value, `openApi.info.version`, `getOperationParameter.ts`'s `parameter.name`) are sanitized before reaching these templates. A single quote in any of them closes the string literal early and the remainder is evaluated as live JS.

## Fix

- New `escapeSingleQuotedString` Handlebars helper (`registerHandlebarHelpers.ts`), registered as a known helper in `rollup.config.mjs`.
- Applied to the six sinks I traced end-to-end to a raw, unsanitized spec field: `path`, `server`, `version`, `prop` (parameter name), request-body `mediaType`, `responseHeader`.
- Deliberately did *not* touch `pattern` (already escaped in `getPattern.ts`) or the enum `name`/`value` fields (already sanitized/pre-escaped in `getEnum.ts`), to avoid double-escaping a value that's already safe.

## Testing

- Re-ran both PoCs from the issue against the fixed build: the injected payload now round-trips as inert string data in the generated output (confirmed by compiling and executing the generated code - no code runs, no marker file gets written).
- `npm test`: 50/50 suites, 80/80 tests, 285/285 snapshots pass, identical to before this change (the new escaping is a no-op for every existing benign fixture).
- `npm run test:e2e`: the browser-driven specs (angular/xhr/fetch/babel) fail identically before and after this change in the sandbox I used ("No usable sandbox", no Chrome sandbox available in this container) - not a regression from this change, every failure is a browser-launch error, not a test assertion failure.

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -38,6 +38,7 @@ const handlebarsPlugin = () => ({
                     enumerator: true,
                     escapeComment: true,
                     escapeDescription: true,
+                    escapeSingleQuotedString: true,
                     camelCase: true,
                 },
             });
--- a/src/templates/client.hbs
+++ b/src/templates/client.hbs
@@ -27,8 +27,8 @@ import { {{{name}}}{{{@root.postfix}}} } from './services/{{{name}}}{{{@root.pos
 		{
 			provide: OpenAPI,
 			useValue: {
-				BASE: OpenAPI?.BASE ?? '{{{server}}}',
-				VERSION: OpenAPI?.VERSION ?? '{{{version}}}',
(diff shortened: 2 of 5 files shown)


carfeii:
Fixes #2809.

